Issue #33, opened 2025-01-14
Address some painpoints in the xpk launcher


tengyifei:
- [x] Don't have to rerun `pip install` if I'm just changing a few files
- [x] Easily select an XPK cluster instead of setting the project/zone/cluster env vars all the time.
- [x] Specify different XLA flags
- [x] Specify the profile/xla_dump bucket.
- [x] Pick a profile directory based on a run name/tag/comment.
- [x] Use different docker URLs so everyone's setup don't collide.
- [ ] Specify an existing docker image URL that contains torchprime, for reproducibility.
- Look into https://skaffold.dev/
- Look into https://k9scli.io/

tengyifei:
>  Pick a profile directory based on a run name/tag/comment.

We can already specify `--name` in `tp run`.

tengyifei:
Filing a separate issue for "Specify an existing docker image URL that contains torchprime, for reproducibility."

PR #43, opened 2025-01-18
Improve xpk workflows

tengyifei:
Tick off some of https://github.com/AI-Hypercomputer/torchprime/issues/33 with this change. We add a `tp` command that will be automatically placed in the user's $PATH when installing the repo. It has two verbs:

- `tp use [...]`: sets the XPK cluster/project/zone and profile/xla dump GCS directory once for all future commands.
- `tp run [...]`: runs a python script on the XPK cluster

In the future, we'll be able to extend more utilities anchoring off the `tp` CLI. For example, we could add a `tp log` that follows the log of a workload, or a `tp bisect` that determines the software version corresponding to a regression.


qihqi (approved):
Nice!

PR #48, opened 2025-01-22
Add a `tp -i` for interactive workflows

tengyifei:
With this change, we can run:

- `tp -i test ...` to run some unit tests
- `tp -i run ...` to run a distributed SPMD command

And the script will monitor the filesystem for changes. It will re-trigger the command again whenever one changes a file, leading to an ergonomic dev/test cycle.

I also found some flags to make the CPU checks faster.

Tested: `tp -i test torchprime/torch_xla_models` to repeatedly run the torch_xla model tests upon edit.


qihqi (approved):

bhavya01:
Thanks! Looks very convenient.

PR #49, opened 2025-01-22
MFU calculator

### LINKED ISSUE #39: Compute throughput and MFU
Rather than manually calculating the step time, throughput, MFU etc, we should add a library to programmatically do that.

tengyifei:
This change adds MFU calculation functionality. It's not wired up to any trainer in any way because we don't have a standard schema for spelling out the LLM layer shapes etc.

Fixes #39.

### DIFF
--- a/torchprime/metrics/mfu.py
+++ b/torchprime/metrics/mfu.py
@@ -1,0 +1,269 @@
+"""
+Calculate Model FLOPs Utilization (MFU).
+"""
+
+import argparse
+import json
+import re
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+
+@dataclass
+class MFU:
+  model_tflops: float
+  """The number of floating point operations in the model, in teraflops."""
+
+  hardware_tflops_per_step: float
+  """The theoretical hardware floating point throughput during one training step, in teraflops."""
+
+  per_chip_tflops_per_sec: float
+  """The realized floating point throughput during one second in each chip, in teraflops."""
+
+  mfu: float
+  """Model FLOPs Utilization. Fraction of hardware FLOPs the model uses, from 0 to 1."""
+
+
+@dataclass
+class Config:
+  per_device_batch_size: int
+  max_target_length: int
+  mlp_dim: int
+  emb_dim: int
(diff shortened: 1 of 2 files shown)


torchprime/metrics/mfu.py
```diff
@@ -0,0 +1,268 @@
...
+  )
+  embedding_flops = (
+    2
+    * config.per_device_batch_size
+    * config.max_target_length
+    * config.emb_dim
+    * config.vocab_size
+  )
+
+  # multiply by 3 for both feed forward and back propagation flops
```
bhavya01, line 108: Multiply by 3 is for gradient checkpointing right?

torchprime/metrics/mfu.py
```diff
@@ -0,0 +1,268 @@
...
+  total_tflops = learnable_weight_tflops + attention_tflops
+
+  if log:
+    print(
+      "Per train step:\n",
+      f"Total TFLOPs: {total_tflops:.2f} \n",
+      f"split as {100 * learnable_weight_tflops/total_tflops:.2f}% learnable weight flops",
+      f"and {100 * attention_tflops/total_tflops:.2f}% attention flops",
+    )
+  return total_tflops, learnable_weight_tflops, attention_tflops
```
bhavya01, line 131: nit: Should we just return total_tflops? or Are the other things returned just for testing?

bhavya01 (approved):

torchprime/metrics/mfu.py
tengyifei: I asked ChatGPT and I believe that's because the backward pass computes two gradients (wrt weights, wrt activations) whereas the forward pass only computes one output (the activation): https://chatgpt.com/share/6792a528-cab0-800b-9ba7-d45718dfaeb3

Based on pure analysis, I think it shouldn't be the case that gradient checkpointing is accounted for in the MFU calculation. Otherwise, we'll be artificially inflating the MFU just by activating checkpointing.

Commented in the code.

torchprime/metrics/mfu.py
tengyifei: Removed. MaxText did this and I just copied it :)

PR #54, opened 2025-01-27
Fix docker URL in xpk

tengyifei:
I forgot to replace the hardcoded pytorch-tpu/llama with the autogenerated docker URL.

Unfortunately in order to return a URL from buildpush, we'll have to convert that from bash to Python.


torchprime/launcher/buildpush.py
```diff
@@ -0,0 +1,81 @@
...
+  docker_url = torchprime_docker_url if torchprime_docker_url else default_url
+
+  print()
+  print(f"Will build a docker image and upload to: {docker_url}")
+  print()
+
+  # Build, tag, and push Docker image
+  try:
+    _run(
+      f"{sudo_cmd} docker build --network=host --progress=auto -t {docker_tag} {context_dir} -f {docker_file}",
```
bhavya01, line 50: Just to confirm, this will rebuild the container when anything in the `context_dir` changes, right?

bhavya01 (approved):

PR #56, opened 2025-01-27
Add a `tp doctor` command

tengyifei:
...and also run this during `tp use`, to check for any missing components/credentials in the environment. This way before the user hit xpk errors in `tp run`, they'll first be guided to fix their environment.

I collected a list of components/credentials to check for while helping @yaochengji get torchprime running.

Example:

```
❯ tp doctor
Checking environment...
Check that docker is installed... ✅
Check that gcloud is logged in... ✅
Check that docker config contains gcr.io credential helper... ✅
Check that the gcloud account can access the gcr.io artifact registry... ✅
Check that kubectl is installed... ✅
Check that gke-gcloud-auth-plugin is installed... ✅
🎉 All checks passed. You should be ready to launch distributed training. 🎉
```

Tested: ideally we'll have end-to-end tests for this but for now I ran these on a new GCE VM.

### DIFF
--- a/torchprime/launcher/doctor.py
+++ b/torchprime/launcher/doctor.py
@@ -68,8 +68,10 @@ def check_docker_access():
     ).stdout.strip()
     raise CheckFailedError(
       f"""The current gcloud account `{account}` cannot access the gcr.io registry.
+The account may not have the required permissions. If it's a service account, the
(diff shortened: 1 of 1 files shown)


yaochengji (approved):
LGTM, thanks!

Issue #19, opened 2025-01-09
[torch_xla_models] Train different Llama variants from our own config files


tengyifei:
Today one could only select a few models on the Huggingface hub. For example, they could pass `meta-llama/Llama-3-8B` to train Llama 3 8B. But I can't e.g. train a single layer llama, or play with the model sizes (e.g. to debug an OOM).

For quick iteration, my suggestion would be the following:
- We could define a few model configurations in `torchprime` itself (e.g. `llama3-8b`, `llama3-70b`, ...) which spell out the layer shapes etc.
- The trainer can take the name of such a config.
- User can locally just write a new model config and test the training.

My 2cents is to try defining the schema with a Python `@dataclass`.

In this issue we don't have to fully solve how to define XLA flags and sharding annotations in the config. We could just pick XLA flags from the env and hardcode to FSDP for now.

tengyifei:
This should be fixed thanks to #57

PR #57, opened 2025-01-28
Replace huggingface args with hydra configs

bhavya01:
Hydra is commonly used config framework that allows composing configs easily. https://hydra.cc/docs/intro/

Adding inital config to allow experimentation with models easily starting with llama. Will extend the config as we support more models and optimizers etc.

Addresses #19


torchprime/torch_xla_models/configs/model/llama.yaml
```diff
@@ -0,0 +1,17 @@
+vocab_size: 128256
```
tengyifei, line 1: Does it make sense to name this file `llama-3-8b.yaml`, since this describes the exact model?

torchprime/torch_xla_models/configs/base.yaml
```diff
@@ -0,0 +1,24 @@
+defaults:
+  - _self_
```
tengyifei, line 2: What does `_self_` mean? Is it possible to add a brief comment?

tengyifei (changes requested):
Could you update the commands in various `README.md` files? Looks like there are about 3 instances when I search for the string `torch_xla_models/` in the repo.

Could we delete `torchprime/torch_xla_models/configs/run.json` now?

tengyifei (approved):

PR #61, opened 2025-01-30
init llama infer

yaochengji:


yaochengji:
@qihqi  could you help take a look whether it's the right way to use torchax for inference?

PR #63, opened 2025-01-30
Add Deepseek-v3

miladm:
Goal:
- [x] Add Deepseek-v3
- [x] Enable single-chip TPU functionality
- [x] Use TorchAx

Non-Goal / Next-Steps:
- [x] Real input tensor
- [x] Real weights
- [x] FP8 quantization kernels enablement (`fp8_gmm`)
- [x] Distrbuted


miladm:
Should probably submit the original model before this PR to easily spot the diff.

qihqi:
Great to get it running! Few high level changes:

1. let's remove all the README.md etc along with their perf graphs, pdfs etc. Instead, in the .py files that we forked, add the URL to their github repo of where we forked from.
2. Make a new README.md with the command you run to run it on TPU
3. edit requirements.txt to TPU requirements.

qihqi (approved):
stamp to unblock feel free to merge after the change

tengyifei:
Can we add a unit test to make sure it runs and produces correct results as compared to CPU? Example for Llama: https://github.com/AI-Hypercomputer/torchprime/blob/main/torchprime/experimental/torchax_models/test/test_llama.py#L40

Btw, we also need to fix lint. You can do that by running `ruff check` and `ruff format`.

Thanks!

yaochengji:
@miladm looks like only part of the model is converted to jax device.

I reverted the code change in model.py and tried to call `model.to("jax")` outside. I got the error: torchax.tensor.OperatorNotFound: Operator with name aten::rms_norm has no lowering.

I can rewrite the rms_norm op to fine grained op as llama model first. @qihqi , BTW, do we apply decomposition before converting torch op to jax ops?

miladm:
I noticed TPU utilization is lower than I anticipated - not suprised if we don't have all tensors on tpu.

@yaochengji I will clean up and merge this PR soon.

torchprime/experimental/torchax_models/DeepSeek-V3/inference/requirements.txt
```diff
@@ -0,0 +1,4 @@
+torch==2.4.1
```
tengyifei, line 1: We should merge these into https://github.com/AI-Hypercomputer/torchprime/blob/main/pyproject.toml#L15 so we don't have a separate set of deps.

torchprime/experimental/torchax_models/DeepSeek-V3/inference/generate.py
```diff
@@ -0,0 +1,214 @@
...
+    total_len = min(model.max_seq_len, max_new_tokens + max(prompt_lens))
+    tokens = torch.full((len(prompt_tokens), total_len), 1, dtype=torch.long, device="cpu") #TODO: bring back -1
+    # for i, t in enumerate(prompt_tokens): #TODO: bring back
+    #     tokens[i, :len(t)] = torch.tensor(t, dtype=torch.long, device="cpu")
+    prev_pos = 0
+    finished = torch.tensor([False] * len(prompt_tokens), device="cpu")
+    prompt_mask = tokens != -1
+
+    env = torchax.default_env()
+    env.to_xla(tokens).jax()
```
qihqi, line 64: this function doesnt do the conversion in-place, but returns a new tensor subclass, so should be `tokens = env.to_xla(tokens)`

miladm:
Finally returning to this PR. Thank you @qihqi for addressing the CI, rough edges.

Issue #67, opened 2025-01-31
Calculate step duration and MFU after a training run is done


tengyifei:
Today we already calculate step duration after a training run. We just need to gather the hardware info (e.g. TPU generation, topology) in order to compute the MFU.

PR #70, opened 2025-01-31
Script to compute step duration

tengyifei:
Precursor to solving #67.

Imports https://github.com/AI-Hypercomputer/tpu-recipes/blob/main/utils/profile_convert.py and improves it.

Specifically, I noticed sometimes there is an empty gap between two step markers in the profile. So if we averaged event durations, that would overestimate the MFU. Instead, this now averages the delta between the starting time offset of neighboring events.

Now that we can print step time from the profile, I removed the step time from the training loop. That added a bunch of delays and is actually pretty inaccurate (1.7s vs 1.85s in local testing).

Tested:

  XLA_IR_DEBUG=1 XLA_HLO_DEBUG=1 python3 torchprime/torch_xla_models/train.py mesh.fsdp=8 profile_step=4 model=llama-3-8b
  XLA_IR_DEBUG=1 XLA_HLO_DEBUG=1 python3 torchprime/torch_xla_models/train.py mesh.fsdp=8 profile_step=4 model=mixtral-8x7b


torchprime/metrics/xplane_pb2.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+# ruff: noqa
+
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
bhavya01, line 5: Is it possible to add a proto file? Not needed if it complicates the build.

bhavya01 (approved):

torchprime/metrics/xplane_pb2.py
tengyifei: It probably requires a custom build step. Let me file an issue.

PR #78, opened 2025-02-03
update README for dependency

zpcore:
Running with `tp` command failed with dependency missing error. We need to install `kubectl`. Mentioning this in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,11 +41,13 @@ Refer to `README.md` in `torchprime/torch_xla_models` and
 
 ### Distributed training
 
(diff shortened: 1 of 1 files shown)


tengyifei:
`tp doctor` already contains a bunch of environment checks, including instructions for installing `kubectl`. WDYT just recommend using that to check dependencies?

zpcore:
Oh nice! We should mention tp doctor before the XPK section. Do you want me update the instruction?

zpcore:
I see `tp run` is calling `tp docker`. I think we can drop the PR.

Issue #22, opened 2025-01-09
Support both FSDP and 2D sharding with the same trainer


tengyifei:
In the huggingface fork, 1D FSDP sharding and 2D FSDP+TP sharding take completely separate code paths. One is via the `FSDPv2` model wrapper and the other is via explicit `mark_sharding` calls on the weights.

In order to enable the application of different model sharding config files, we should first explore how to support both 1D and 2D sharding with the same code path.

For example, maybe we could design a `ShardingConstraints` model wrapper just like FSDP, and have it:

- Match the PyTorch module class by name
- Shard the inputs according to the config
- Shard the outputs according to the config
- Shard the weights according to some wildcard matching
- And possibly shard gradients during the backward pass

This should be general enough to support all sorts of sharding except pipeline parallelism.

Issue #21, opened 2025-01-09
Support changing the sharding from a model config


tengyifei:
Once it's possible to configure the shapes/rope embeddings/norms of a model, we should also support configuring the sharding of various parameters from a config file. The config file could be in Python or in `.yaml` etc.

One idea for reference: https://github.com/AI-Hypercomputer/jetstream-pytorch/blob/main/default_shardings/llama.yaml

The expectation is that someone won't specify `--fsdp` or `--spmd_2d_sharding` on the CLI. Instead, they can specify a sharding config name/path. We'll expect to implement FSDP sharding via the config. And we'll be able to test a variety of sharding configs.

bhavya01:
We also need to add shardings to intermediate tensors sometimes to model code. It is harder to do this with a sharding config without named tensors https://github.com/pytorch-tpu/transformers/blob/expert_parallel/src/transformers/models/mixtral/modeling_mixtral.py#L871C31-L872C13

PR #84, opened 2025-02-05
Start of a generic config based model sharder

tengyifei:
## Motivation

In Hugging Face transformer models, we have been using FSDPv2 [1] and hardcoded `mark_sharding` calls to annotate tensors with sharding constraints. This is not very extensible. FSDPv2 doesn't support 2D sharding, and changing sharding config requires modifying the model code.

This PR introduces the start of a config based model sharder that generalizes to arbitrary sharding schemes supported by GSPMD, e.g. 2D FSDP+TP sharding, expert parallelism, etc. It adds a `shard_model_from_config` function that can take a vanilla single-device PyTorch model, then replace the weights and activations with sharded versions, without code change.

## Design

The user can specify the names of weights to shard and their partition spec. They can also shard certain activations by specifying the names of the `nn.Module` that outputs that activation. Here's an example for Llama 2D sharding:

  ```yaml
  # Weights
  model.embed_tokens.weight: ['fsdp', null]
  model.layers.*.self_attn.q_proj.weight: ['fsdp', null]
  model.layers.*.self_attn.k_proj.weight: [null, 'fsdp']
  model.layers.*.self_attn.v_proj.weight: [null, 'fsdp']
  model.layers.*.self_attn.o_proj.weight: ['fsdp', null]
  model.layers.*.mlp.gate_proj.weight: ['fsdp', null]
  model.layers.*.mlp.up_proj.weight: ['fsdp', null]
  model.layers.*.mlp.down_proj.weight: [null, 'fsdp']
  model.layers.*.input_layernorm.weight: ['fsdp']
  model.layers.*.post_attention_layernorm.weight: ['fsdp']
  model.norm.weight: ['fsdp']
  lm_head.weight: ['fsdp', null]

  # Activations
  model.embed_tokens: ['fsdp', null, null]
  model.layers.*.self_attn: ['fsdp', null, null]
  model.layers.*: ['fsdp', null, null]
  model.lm_head: ['fsdp', null, null]
  ```

The `*` will match to any integer name, which is useful for a list of layers.

In this PR, they can shard a model by passing this config to `shard_model_from_config` in Python. In future PRs, we'll support sharding a model from a yaml configuration file as well.

Under this scheme, it's immediately obvious how is a particular param or activation sharded. We can later create different configs for FSDP, 2D sharding, and so on.

To shard parameters, we traverse the state dict and call mark_sharding.

To shard activations, we replace the matching `nn.Module` with a `ShardedModule` wrapper that runs the underlying module, and then call `mark_sharding` to shard the output.

The `mark_sharding` is a pluggable callback so that we could use this with both `torch_xla` and `torchax`.

Part of #21 and #22.

[1]: https://github.com/pytorch/xla/issues/6379

### DIFF
--- a/torchprime/sharding/shard_model.py
+++ b/torchprime/sharding/shard_model.py
@@ -1,0 +1,178 @@
+from collections.abc import Callable
+
+import torch.nn
+
+ShardWeightFn = Callable[[torch.Tensor, str], torch.Tensor]
+"""
+ShardWeightFn optionally transforms a weight tensor based on its name.
+
+Args:
+
+  weight (torch.Tensor): The weight tensor to be transformed.
+  
+  name (str): The name of the weight tensor as it appears in the state dict.
+
+Returns:
+
+  torch.Tensor: The transformed weight tensor.
+
+"""
+
+
+ShardActivationFn = Callable[[torch.nn.Module, str], torch.nn.Module]
+"""
+ShardActivationFn optionally transforms a module based on its name.
+
+Args:
+
+  module (torch.nn.Module): The module to be transformed.
+  
+  name (str): The name of the module as it appears in the state dict.
+
+Returns:
+
+  torch.nn.Module: The transformed module, or the original module if
+    no transformation is needed.
+
+"""
+
+
+def shard_model(
+  model: torch.nn.Module,
+  shard_weight: ShardWeightFn,
+  shard_activation: ShardActivationFn,
+) -> torch.nn.Module:
+  """
+  Transforms `model` by applying `shard_weight` to each weight tensor and applying
+  `shard_activation` to each module. Returns the transformed module.
+  """
+  state_dict = {}
+  for name, param in model.state_dict().items():
+    state_dict[name] = shard_weight(param, name)
+  model.load_state_dict(state_dict, assign=True)
+  return _wrap_module(model, shard_activation, tuple())
+
+
+def _wrap_module(
+  mod: torch.nn.Module, shard_activation: ShardActivationFn, prefix: tuple[str, ...]
+) -> torch.nn.Module:
+  """
+  Recursively transforms the modules to shard activations.
+
+  Start from the leaf modules and work our way up, to handle cases where one
+  module is the child of another. The child modules will be transformed first,
+  and then the parent module will be transformed, possibly with transformed
+  children.
+  """
+  new_children = {}
+  for name, child in mod.named_children():
+    new_children[name] = _wrap_module(child, shard_activation, prefix + (name,))
+  for name, new_child in new_children.items():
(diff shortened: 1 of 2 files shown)


torchprime/sharding/shard_model.py
```diff
@@ -0,0 +1,173 @@
...
+  """
+  new_children = {}
+  for name, child in mod.named_children():
+    new_children[name] = _wrap_module(child, shard_activation, prefix + (name,))
+  for name, new_child in new_children.items():
+    mod.set_submodule(name, new_child)
+  return shard_activation(mod, ".".join(prefix))
+
+
+def shard_model_from_config(
```
bhavya01, line 75: Do you think it'll be easier if we just accept a mesh argument and pass the mesh argument to shard_output and shard_param callables?

torchprime/sharding/tests/test_shard_model.py
```diff
@@ -0,0 +1,237 @@
...
+
+  model = shard_model(model, lambda x, _: x, shard_activation)
+  assert visited == {"", "fc1", "relu", "fc2"}
+
+  # Check that all modules are sharded.
+  sharded = set()
+  for name, mod in model.named_modules():
+    if isinstance(mod, MockShardedModule):
+      sharded.add(name)
+  assert len(sharded) == len(visited)
```
bhavya01, line 82: Wasn't expecting an empty name in the list while going over named_modules but good to know.

torchprime/sharding/tests/test_shard_model.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def test_shard_model_from_config_mock():
+  model = nn.Sequential(SimpleLinear(), SimpleLinear())
+  config = {
```
bhavya01, line 116: To specify that we want to shard the whole model output, would we need just an empty string with partition spec?

tengyifei: Indeed! And it should work in the yaml format too

torchprime/sharding/tests/test_shard_model.py
tengyifei: Commented

torchprime/sharding/shard_model.py
tengyifei: That would be convenient but I wanted this to work across `torchax` and `torch_xla`, which takes different meshes. I think it's still useful to have a convenience wrapper around this, so I'll add a TODO here to address in the next PR.

tengyifei: Filed https://github.com/AI-Hypercomputer/torchprime/issues/86

bhavya01 (approved):

PR #95, opened 2025-02-07
support run with huggingface models

zpcore:
Support run huggingface models with the following command:
```
tp run --use-hf torchprime/hf_models/train.py
```
HF model config files are placed under torchprime/hf_models/... 

Other small update:
- Pin XPK dependency to `xpk@git+https://github.com/AI-Hypercomputer/xpk@e52a5f4cd56ad50aeab06d55100cf4d3abc4c2c8`


torchprime/hf_models/configs/default.yaml
```diff
@@ -0,0 +1,36 @@
+defaults:
+  - _self_
+
+# extra env variables you want to pass to the training script
+env:
+  - NUM_SLICE: 1
+  - XLA_PERSISTENT_CACHE_PATH: "/tmp/xla_cache"
+  - PROFILE_LOGDIR: "gs://dummy-bucket"
```
miladm, line 8: can we have a better naming convtion?

zpcore: This needs the user to config where to store the profile. Updated with a better naming.

torchprime/launcher/cli.py
```diff
@@ -210,20 +211,27 @@ def run(args):
...
+    ] + list(args)
+  else:
+    command = ["python", "torchprime/launcher/thunk.py"] + list(args)
 
   # Forward a bunch of important env vars.
   env_forwarding = [
     *forward_env("HF_TOKEN"),  # HuggingFace token
     *forward_env("XLA_IR_DEBUG"),  # torch_xla debugging flag
     *forward_env("XLA_HLO_DEBUG"),  # torch_xla debugging flag
+    *forward_env("LIBTPU_INIT_ARGS"),  # XLA flag
```
tengyifei, line 234: Can we update the readme to mention that `LIBTPU_INIT_ARGS` is forwarded, next to the other three flags

torchprime/torch_xla_models/configs/model/llama-3-70b.yaml
```diff
@@ -0,0 +1,27 @@
+model_class: llama.LlamaForCausalLM # Used to import the model from this class
```
tengyifei, line 1: This should be split to a separate PR since it doesn't apply to huggingface.

torchprime/hf_models/configs/default.yaml
```diff
@@ -0,0 +1,35 @@
+defaults:
+  - _self_
+
+# extra env variables you want to pass to the training script
+env:
+  - NUM_SLICE: 1
```
tengyifei, line 6: Is `NUM_SLICE` needed by anything?

torchprime/launcher/cli.py
```diff
@@ -210,20 +211,27 @@ def run(args):
...
   if docker_project is None:
     docker_project = config.project
-  docker_url = buildpush(docker_project)
+  docker_url = buildpush(docker_project, build_arg=build_arg)
 
   # Submit xpk workload
   datetime_str = datetime.now().strftime("%Y%m%d-%H%M%S")
-  command = ["python", "torchprime/launcher/thunk.py"] + list(args)
+  if use_hf:
+    command = [
```
tengyifei, line 223: I think it's better to have the HF train command wrapped by `thunk.py`. The thunk does a few things:

- Mount a google storage bucket to store profiles
- Also configure XLA_FLAGS to store HLO graph in the bucket
- Append the user's preferred artifact dir as a `profile_dir=...` CLI argument

That's how all other workload is invoked and stores their profiles.

Then we also don't need a separate PROFILE_LOGDIR config.

tengyifei (changes requested):

torchprime/hf_models/configs/default.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - XLA_PERSISTENT_CACHE_PATH: "/tmp/xla_cache"
+  - PROFILE_LOGDIR: "gs://torchprime/profile"
```
tengyifei, line 8: See my comment at [1], we don't need to separately configure the profile dir here.

[1]: https://github.com/AI-Hypercomputer/torchprime/pull/95#discussion_r1947049868

torchprime/hf_models/configs/default.yaml
```diff
@@ -0,0 +1,35 @@
+defaults:
+  - _self_
+
+# extra env variables you want to pass to the training script
+env:
+  - NUM_SLICE: 1
+  - XLA_PERSISTENT_CACHE_PATH: "/tmp/xla_cache"
```
tengyifei, line 7: I used to also do this but it causes the compiler to generate ~1GB of graph cache for every graph. The cache is then destroyed at the end of training since it's mounted at `/tmp`. I think this might even slow things down for no obvious benefit so we could probably remove it..

torchprime/hf_models/configs/default.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - PROFILE_LOGDIR: "gs://torchprime/profile"
+  - XLA_USE_SPMD: 1
+  - ENABLE_PJRT_COMPATIBILITY: true
```
tengyifei, line 10: Is the `ENABLE_PJRT_COMPATIBILITY` flag needed by anything? From what I could tell, it's specific to JAX and not PyTorch/XLA.

torchprime/hf_models/configs/default.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - PROFILE_DURATION_MS: 100000
+
```
tengyifei, line 12: We could add a `profile_dir` config similar to https://github.com/AI-Hypercomputer/torchprime/blob/main/torchprime/torch_xla_models/configs/default.yaml#L21 -- this value will become what the user sets in `tp use`, followed by some date and slice ID folders etc.

torchprime/hf_models/train.py
```diff
@@ -0,0 +1,53 @@
...
+def main(config: DictConfig) -> None:
+  try:
+    set_env_variables(config)
+    hf_token = os.environ.get("HF_TOKEN")
+    if hf_token:
+      subprocess.run(["huggingface-cli", "login", "--token", hf_token], check=True)
+
+    # Build and run command
+    cmd = build_command(config)
+    subprocess.run(cmd, check=True)
```
tengyifei, line 46: This seems like good place to add more result postprocessing in the future.

torchprime/launcher/Dockerfile
```diff
@@ -43,4 +44,18 @@ COPY . /workspaces/torchprime
 # This should not install any packages. Only symlink the source code.
 RUN pip install --no-deps -e .
 
+RUN if [ "$USE_TRANSFORMERS" = "true" ] && [ -d "transformers" ]; then \
```
tengyifei, line 47: I'm not sure this works, since we configure docker build to ignore all `.gitignore` files. Then `transformers` will never be part of the build context here and it might forever use the GitHub version.

torchprime/launcher/Dockerfile
```diff
@@ -43,4 +44,18 @@ COPY . /workspaces/torchprime
...
+        echo "Using local transformers repo"; \
+        cp -r transformers /workspaces/transformers; \
+    elif [ "$USE_TRANSFORMERS" = "true" ]; then \
+        echo "Cloning transformers from GitHub"; \
+        git clone --depth 1 https://github.com/huggingface/transformers.git /workspaces/transformers; \
+    fi
+
+# Only install transformers if USE_TRANSFORMERS is true
+RUN if [ "$USE_TRANSFORMERS" = "true" ]; then \
+        pip install --no-deps -e /workspaces/transformers; \
```
tengyifei, line 57: Did you test that transformer trains okay? If we install with the `--no-deps` flag, it might be missing a bunch of dependencies.

README.md
```diff
@@ -120,6 +120,17 @@ and attributes where this model code came from, if any. This also helps to
 show case what changes we have done to make it performant on TPU. The original
 version is not expected to be run.
 
+## Run huggingface transformer models
+Torchprime supports run with huggingface models by taking advantage of `tp run`. To use huggingface models, you can clone [huggingface/transformers](https://github.com/huggingface/transformers) under torchprime. This allows you to pick any branch or make code modifications for experiment:
```
tengyifei, line 124: nit: the rest of this file is wrapped at 80 characters for readability. I use a vscode extension called "rewrap" to do this automatically.

tengyifei (changes requested):

torchprime/launcher/cli.py
zpcore: Agree on using thunk.py to specify profile dir. One thing to notice is that `--profile_dir` may not be an argument for `..trainer.py`, need to reorganize it as an env or argument.

tengyifei: Yeah, you just need to add a `profile_dir: profile` field to the default config yaml like we do for other yamls. Also note that the CLI syntax used by thunk is `profile_dir=...` instead of `--profile_dir=...`

zpcore: Oh, I see, you basically overwrite the key-value in config.

torchprime/launcher/Dockerfile
zpcore: Yes, it works with `--no-deps`(just miss evaluate). I don't want transformers to reinstall torch/torch_xla here accidently.

torchprime/hf_models/configs/default.yaml
zpcore: Was for backward compatibility. Remove it should be fine now.

torchprime/hf_models/configs/default.yaml
zpcore: Yes, for XPK run, this is not useful. Removed.

torchprime/launcher/Dockerfile
zpcore: Use seperate .gitignore and .dockerignore for now.

tengyifei: Ack

torchprime/launcher/Dockerfile
tengyifei: Ack

torchprime/torch_xla_models/configs/model/llama-3-70b.yaml
tengyifei, line 1: Let's remove this file since it is not relevant to this PR

tengyifei (approved):

torchprime/torch_xla_models/configs/model/llama-3-70b.yaml
zpcore: Ops, removed.

PR #106, opened 2025-02-11
Do not terminate run when running out of sample

zpcore:


torchprime/torch_xla_models/train.py
```diff
@@ -206,29 +200,32 @@ def train_loop(self):
...
 
+    epoch = 0
     for step in range(max_step):
       try:
         batch = next(train_iterator)
       except StopIteration:
-        break
+        logger.warning(f"DataLoader exhausted at step {step}, reset iterator")
+        epoch += 1
+        train_iterator = iter(train_loader)
```
tengyifei, line 210: Did you test training for more than 1 epochs? From the look of this code, the `batch` variable will be uninitialized if we get here via an exception.

I would've thought we needed something like

```python
try:
  batch = next(train_iterator)
except StopIteration:
  logger.warning(f"DataLoader exhausted at step {step}, reset iterator")
  epoch += 1
  train_iterator = iter(train_loader)
  batch = next(train_iterator)
```

It's also good to factor this out into a separate object

tengyifei (changes requested):

torchprime/torch_xla_models/train.py
zpcore: I have the test here for more than one epoch [link](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22tpu-pytorch%22%0Aresource.labels.location%3D%22us-east1%22%0Aresource.labels.cluster_name%3D%22pytorch-v6e-256-donotdelete-uj%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.pod_name:%22piz-xpk-v6e-256-1-20250211-073156-slice-job-0-0-%22%20severity%3E%3DDEFAULT%0ASEARCH%2528%22loss%22%2529;storageScope=project;cursorTimestamp=2025-02-11T07:53:19.650241358Z;duration=P1D?e=13802955&mods=allow_workbench_image_override&project=tpu-pytorch&inv=1&invt=AbpRSg), but I think you are right. Let me add it.

PR #107, opened 2025-02-12
Add a common rope impl

tengyifei:
This rope impl supports the OG rope and the scaling used in Llama 3.1

### DIFF
--- a/torchprime/rope/rope.py
+++ b/torchprime/rope/rope.py
@@ -1,0 +1,74 @@
+"""
+Rotary Positional Embeddings (RoPE) implementation.
+Reference: https://github.com/adalkiran/llama-nuts-and-bolts/blob/main/docs/10-ROPE-ROTARY-POSITIONAL-EMBEDDINGS.md
+"""
+
+import math
+from dataclasses import dataclass
+
+import torch
+
+
+@dataclass
+class RopeScaling:
+  """
+  RoPE scaling parameters. The defaults are what was selected in Llama 3.1.
+  """
+
+  factor: float = 8.0
+  low_freq_factor: float = 1.0
+  high_freq_factor: float = 4.0
+  original_context_len: int = 8192
+
+
+def default_rope_frequencies(
+  head_dim: int,
+  theta: float = 10000.0,
+) -> torch.Tensor:
+  """
+  Computes the original RoPE frequencies in e.g. Llama 2.
+  Args:
+      head_dim: the size of a single attention head.
+      theta: a hyperparameter controlling how fast the embeddings rotate.
+  Returns:
+      The frequencies for the RoPE embeddings.
+  """
+  return 1.0 / (
+    theta ** (torch.arange(0, head_dim, 2, dtype=torch.int64).float() / head_dim)
+  )
+
+
+def llama3_rope_frequencies(
+  head_dim: int,
+  theta: float = 10000.0,
+  scaling: RopeScaling | None = None,
+) -> torch.Tensor:
+  """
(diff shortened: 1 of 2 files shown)


torchprime/tests/test_rope.py
```diff
@@ -0,0 +1,120 @@
...
+  )
+
+  assert scale == 1
+  torch.testing.assert_close(ours, hf)
+
+
+@pytest.mark.parametrize(
+  "hidden_size, num_attention_heads, theta",
+  [(4096, 32, 500000.0), (16384, 128, 500000.0), (65536, 128, 500000.0)],
+)
```
pgmoka, line 76: context: I am new to `@pytest`. I am confused by this call. Is this necessary given lines 11

pgmoka (commented):
Mainly nits and questions on tests

torchprime/rope/__init__.py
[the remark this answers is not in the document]
tengyifei: I think it's a Pythoni-ism: https://stackoverflow.com/questions/448271/what-is-init-py-for 
basically if we don't have this file, then some import fails.

torchprime/rope/rope.py
```diff
@@ -0,0 +1,65 @@
...
+
+import torch
+
+
+@dataclass
+class RopeScaling:
+  factor: float = 8.0
+  low_freq_factor: float = 1.0
+  high_freq_factor: float = 4.0
+  original_context_len: int = 8192
```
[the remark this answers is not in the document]
tengyifei: Done

torchprime/tests/test_rope.py
tengyifei: This is just the Python testing way of declaring a table driven test for each test case.

I'll factor the code to share those params.

torchprime/tests/test_rope.py
```diff
@@ -0,0 +1,120 @@
...
+def test_llama3_rope_against_hf(hidden_size, num_attention_heads, theta):
+  head_dim = hidden_size // num_attention_heads
+  ours = rope.llama3_rope_frequencies(
+    head_dim=head_dim,
+    theta=theta,
+    scaling=rope.RopeScaling(
+      factor=8,
+      low_freq_factor=1,
+      high_freq_factor=4,
+      original_context_len=8192,
```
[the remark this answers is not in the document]
tengyifei: I factored this into a global. AFAIK, only Llama 3 use this kind of scaling params, and they only use this specific set of values. So it should be fine to just test this case.

torchprime/rope/rope.py
```diff
@@ -0,0 +1,69 @@
...
+  if scaling is None:
+    return freqs
+
+  low_freq_wavelen = scaling.original_context_len / scaling.low_freq_factor
+  high_freq_wavelen = scaling.original_context_len / scaling.high_freq_factor
+
+  wavelen = 2 * math.pi / freqs
+  # wavelen < high_freq_wavelen: do nothing
+  # wavelen > low_freq_wavelen: divide by factor
+  freqs = torch.where(wavelen > low_freq_wavelen, freqs / scaling.factor, freqs)
```
bhavya01, line 60: Should we add a warning if low_freq_wavelen is smaller than high_freq_wavelen? In that case, this might deviate from original implementation?

bhavya01 (approved):

torchprime/rope/rope.py
tengyifei: *nod* Made it an assertion since I think that would be very surprising and we can still relax it if needed.

PR #108, opened 2025-02-12
Initial Llama 3.1 405B support

tengyifei:
Adds Llama 3.1 405B single pod support for torch_xla.

In order to train 405B on a Trillium pod, we need 64 FSDP by 4 TP
sharding. I added a `llama-fsdp-tp` scaling config that expresses the 2D
sharding configuration and we can also use that for other llama models.

This trains performantly on 1 Trillium pod. The step time was 27.395s
(see http://shortn/_fkc7ggB8PE), which is on par with the
Huggingface result (27.395s).

Longer run over 500 steps: http://shortn/_wyNRllBbOA

Tested: `tp run torchprime/torch_xla_models/train.py model=llama-3.1-405b global_batch_size=64 mesh.fsdp=64 mesh.tensor=4 dataset_config_name=wikitext-103-raw-v1 profile_step=15 profile_duration=100000 max_steps=500 logging_steps=10`

### DIFF
--- a/torchprime/torch_xla_models/configs/model/scaling/llama-fsdp-tp.yaml
+++ b/torchprime/torch_xla_models/configs/model/scaling/llama-fsdp-tp.yaml
@@ -1,0 +1,27 @@
+# 2D (FSDP + TP) sharding configuration for Llama models.
+
+activation_checkpoint_layers:
+ - LlamaDecoderLayer
+
+# Refer to https://github.com/pytorch/xla/issues/6379 for backward optimization barrier info.
+optimization_barrier_layers:
+ - LlamaDecoderLayer
+
+sharding:
+  # Weights
+
+  # TODO(https://github.com/AI-Hypercomputer/torchprime/issues/114): This
+  # cannot be `[tensor, fsdp]`, or the gradients will sometimes become NaN.
+  model.embed_tokens.weight: [fsdp, tensor]
+
+  model.layers.*.self_attn.q_proj.weight: [tensor, fsdp]
+  model.layers.*.self_attn.k_proj.weight: [tensor, fsdp]
+  model.layers.*.self_attn.v_proj.weight: [tensor, fsdp]
+  model.layers.*.self_attn.o_proj.weight: [fsdp, tensor]
+  model.layers.*.mlp.gate_proj.weight: [tensor, fsdp]
+  model.layers.*.mlp.up_proj.weight: [tensor, fsdp]
+  model.layers.*.mlp.down_proj.weight: [fsdp, tensor]
+  model.layers.*.input_layernorm.weight: [fsdp]
+  model.layers.*.post_attention_layernorm.weight: [fsdp]
+  model.norm.weight: [fsdp]
+  lm_head.weight: [tensor, fsdp]
--- a/torchprime/torch_xla_models/llama/model.py
+++ b/torchprime/torch_xla_models/llama/model.py
@@ -27,6 +27,7 @@ from torch import nn
 from transformers.activations import ACT2FN
 from transformers.utils import logging
 
+from torchprime.rope.rope import RopeScaling, llama3_rope_frequencies
 from torchprime.torch_xla_models.loss import cross_entropy_loss
 
 logger = logging.get_logger(__name__)
@@ -50,18 +51,16 @@ class LlamaRMSNorm(nn.Module):
 
 
 class LlamaRotaryEmbedding(nn.Module):
+  inv_freq: nn.Buffer
+
   def __init__(
-    self, dim, max_position_embeddings=2048, base=10000, device=None, scaling_factor=1.0
+    self,
+    head_dim,
+    rope_theta,
+    scaling: RopeScaling | None = None,
   ):
     super().__init__()
-    self.scaling_factor = scaling_factor
-    self.dim = dim
-    self.max_position_embeddings = max_position_embeddings
-    self.base = base
-    inv_freq = 1.0 / (
-      self.base
-      ** (torch.arange(0, self.dim, 2, dtype=torch.int64).float().to(device) / self.dim)
-    )
+    inv_freq = llama3_rope_frequencies(head_dim, theta=rope_theta, scaling=scaling)
     self.register_buffer("inv_freq", inv_freq, persistent=False)
 
   @torch.no_grad()
@@ -203,10 +202,11 @@ class LlamaAttention(nn.Module):
     self._init_rope()
 
   def _init_rope(self):
+    scaling = self.config.get("rope_scaling", None)
+    if scaling is not None:
+      scaling = RopeScaling(**scaling)
     self.rotary_emb = LlamaRotaryEmbedding(
-      self.head_dim,
-      max_position_embeddings=self.max_position_embeddings,
-      base=self.rope_theta,
+      head_dim=self.head_dim, rope_theta=self.rope_theta, scaling=scaling
     )
 
   @xp.trace_me("LlamaAttention")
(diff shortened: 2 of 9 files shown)


torchprime/torch_xla_models/configs/model/scaling/llama-fsdp-tp.yaml
```diff
@@ -0,0 +1,27 @@
...
+  model.layers.*.self_attn.k_proj.weight: [tensor, fsdp]
+  model.layers.*.self_attn.v_proj.weight: [tensor, fsdp]
+  model.layers.*.self_attn.o_proj.weight: [fsdp, tensor]
+  model.layers.*.mlp.gate_proj.weight: [tensor, fsdp]
+  model.layers.*.mlp.up_proj.weight: [tensor, fsdp]
+  model.layers.*.mlp.down_proj.weight: [fsdp, tensor]
+  model.layers.*.input_layernorm.weight: [fsdp]
+  model.layers.*.post_attention_layernorm.weight: [fsdp]
+  model.norm.weight: [fsdp]
+  lm_head.weight: [tensor, fsdp]
```
zpcore, line 27: Do we need to specify the activation sharding?

tengyifei: In this version, I didn't specify any activation sharding, to match exactly the behavior of https://github.com/pytorch-tpu/transformers/blob/d28fc1229a38c328f53cdad0063b67b370461ca8/examples/pytorch/language-modeling/run_clm.py#L548. Just like the `llama-fsdp.yaml` replicating the behavior of the `FSDPv2` wrapper.

I just learned from Blake that this implements what's called "megatron-style" 2d sharding. There's another variation of 2d sharding that requires sharding the activations, and that could save some RAM and enable collective matmul. That's something I plan to look into later.

torchprime/torch_xla_models/tests/test_llama.py
```diff
@@ -10,95 +13,174 @@
...
+  )
+  # Place model on CPU device first
+  with torch.device("cpu"):
+    hf_model = HfLlamaForCausalLM(config)
+    model = LlamaForCausalLM(torchprime_config)
+    # Assert that the `inv_freq` values are the same
+    assert isinstance(model.model.layers[0].self_attn, nn.Module)
+    assert isinstance(hf_model.model.layers[0].self_attn, nn.Module)
+    assert isinstance(model.model.layers[0].self_attn.rotary_emb, nn.Module)
+    assert isinstance(hf_model.model.layers[0].self_attn.rotary_emb, nn.Module)
```
zpcore, line 107: Curious are those assert necessary?

tengyifei: Technically they're not necessary, but if you're writing in an IDE, this helps the language server/code autocomplete figure out the type of the module and suppresses some type check errors.

tengyifei:
Here is a longer run showing loss decrease of 500 steps: http://shortn/_KCyeI2Gr61

torchprime/torch_xla_models/configs/model/scaling/llama-fsdp-tp.yaml
zpcore: HF run `spmd_2d_sharding` is not enabled in the test config. Did I miss anything?

torchprime/torch_xla_models/train.py
```diff
@@ -48,6 +49,31 @@
...
+  slice_indices = []
+  for d in device_attributes:
+    index = d.get("slice_index", "0")
+    assert isinstance(index, str), f"Expected slice_index to be a string, got {index}"
+    slice_indices.append(int(index))
+  num_slices = max(slice_indices) + 1
+  return num_slices > 1
+
+
+def _is_1d_sharding(mesh_values: tuple[int, ...]) -> bool:
```
bhavya01, line 66: Would it be better if we moved these functions to a utils file and added tests for it. That protects us from any changes that happens to these torch_xla api

torchprime/torch_xla_models/train.py
```diff
@@ -305,14 +360,13 @@ def main(config: DictConfig):
...
+  # Set the model dtype to bfloat16
+  with set_default_dtype(torch.bfloat16), torch_xla.device():
+    model = initialize_model_class(config.model)
+
   n_params = sum([p.numel() for p in model.parameters()])
   logger.info(f"Training new model from scratch - Total size={n_params} params")
 
-  # Set the model dtype to bfloat16
-  model = model.to(torch.bfloat16)
-  model = model.to("xla")
```
bhavya01, line 314: nit: Is the model initialized directly on the device?

torchprime/torch_xla_models/llama/model.py
```diff
@@ -203,10 +202,11 @@ def __init__(self, config: DictConfig, layer_idx: int | None = None):
     self._init_rope()
 
   def _init_rope(self):
+    scaling = self.config.get("rope_scaling", None)
+    if scaling is not None:
+      scaling = RopeScaling.from_dict(scaling)  # type: ignore
```
bhavya01, line 207: nit: Is it possible to initialize like to avoid dataclasses-json dependency:

```
scaling = RopeScaling(**scaling)
```
Ref: https://stackoverflow.com/questions/68417319/initialize-python-dataclass-from-dictionary

torchprime/torch_xla_models/train.py
tengyifei: Yes, now the model is initialized directly on the TPU. For smaller models this shouldn't affect things. For large models like Llama 3.1 405B, it's actually impossible to initialize the model on the CPU (due to insufficient host memory) so the TPU route is the only option, short of hacks like initializing every layer manually.

tengyifei: For Llama 405B on 2 pods, I had to add the `--xla_tpu_spmd_rng_bit_generator_unsafe=true` XLA flag to enable sharding propagation on the RNG. For smaller models, we could probably do without this flag.

bhavya01: I am a little confused regarding how that happens. initialize_model_class initialize the model and none of the weight initializations here specify the device. It seems like the sharding code moves the tensors to the device. Can you provide some clarity here?

https://github.com/AI-Hypercomputer/torchprime/blob/main/torchprime/torch_xla_models/llama/model.py#L200

tengyifei: Yes. The `with set_default_dtype(torch.bfloat16), torch_xla.device():` context manager does two things:
- Set the default dtype to bf16
- Set the default device to the XLA device

Under this context manager, all the weights will be initialized on the XLA device. When we run `nn.Linear(...)` under the `with torch_xla.device():` context manager, the `nn.Linear(...)` constructor gets captured into LazyTensor graph. No actual initialization takes place. This way, we can `xs.mark_sharding` individual weights to specify _post-hoc_ that those weights get sharded/placed in a specific way. Finally, a `torch_xla.sync()` call executes this whole graph. In summary, this is a very clean way to initialize the model weights in a sharded way by capturing the model constructor into HLO.

torchprime/torch_xla_models/train.py
tengyifei: Yes. I couldn't test `is_multi_slice` super extensively because we don't know what's the environment these tests will be run under, but at least I could add some tests and make sure they don't crash.

tengyifei: Done

torchprime/torch_xla_models/train.py
bhavya01: Ah, makes sense!

bhavya01 (approved):

tengyifei:
For posterity I extracted the loss from the 500 step run in this PR before the logs are permanently gone: http://shortn/_1LAxT88dNO

Loss vs  Step

PR #109, opened 2025-02-12
rename transformers folder to avoid naming conflit with package

zpcore:
Previously we clone transformers/ under torchprime in order to run huggingface models. This caused the issue that python files who `import transformers` will consider transformers as the one from local folder.

We renaming the `transformers` --> `local_transformers` so that it won't conflict with the package naming.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ htmlcov/
 
 # Ignore local transformers repo
 **/transformers/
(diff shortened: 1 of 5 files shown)


tengyifei (approved):
Thanks!

zpcore:
I am launching a run to confirm the fix.

PR #110, opened 2025-02-13
Update HF runner configs

tengyifei:
Added a Llama 3.1 405B configuration.

Also fixed some comments in the config yaml.
In particular for per_device_train_batch_size, I based the comment on http://shortn/_SR8wqscDlo.

Also changes Dockerfile to install all transformer deps, because the local_transformer may be newer than our transformer dependency. In that case it may require a newer dependency such as tokenizers==0.20 instead of tokenizers==0.19.

### DIFF
--- a/torchprime/launcher/Dockerfile
+++ b/torchprime/launcher/Dockerfile
@@ -52,8 +52,7 @@ RUN if [ "$USE_TRANSFORMERS" = "true" ] && [ -d "local_transformers" ]; then \
 
 # Only install transformers if USE_TRANSFORMERS is true
 RUN if [ "$USE_TRANSFORMERS" = "true" ]; then \
-        pip install --no-deps -e /workspaces/torchprime/local_transformers; \
-        pip install --no-deps evaluate; \
+        pip install -e /workspaces/torchprime/local_transformers evaluate; \
     fi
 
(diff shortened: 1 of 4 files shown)


torchprime/launcher/Dockerfile
```diff
@@ -52,8 +52,7 @@ RUN if [ "$USE_TRANSFORMERS" = "true" ] && [ -d "local_transformers" ]; then \
 
 # Only install transformers if USE_TRANSFORMERS is true
 RUN if [ "$USE_TRANSFORMERS" = "true" ]; then \
-        pip install --no-deps -e /workspaces/torchprime/local_transformers; \
-        pip install --no-deps evaluate; \
+        pip install -e /workspaces/torchprime/local_transformers evaluate; \
```
zpcore, line 55: My concern is that this will overwrite the `torch` version in the base docker image. Did you see any issue with `--no-deps`?

zpcore (approved):

torchprime/launcher/Dockerfile
tengyifei: Yeah, the issue I had was that sometimes the `transformer` local checkout can be newer, and requires a newer `tokenizer==0.20` dependency. This way, we'll always ensure that the required deps are installed.

As to the `torch` version issue, the deps requirement on HF [1] seems pretty relaxed. So it will happily used the existing torch (from ptxla docker).

[1]: https://github.com/huggingface/transformers/blob/main/setup.py#L185-L187

PR #111, opened 2025-02-13
Support run trainer locally

zpcore:
Support to run train.py locally:
```
tp dbrun --use-hf torchprime/hf_models/train.py # huggingface run
```
Or 
```
tp dbrun  torchprime/torch_xla_models/train.py 
```
This makes it easier to debug model code.

Currently it is still using the torch_xla from the base docker image. In the future, we may want to give the option to run with local build torch_xla.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,6 @@ Finally, each model may also provide a GPU "original" version that illustrates
 and attributes where this model code came from, if any. This also helps to
 show case what changes we have done to make it performant on TPU. The original
 version is not expected to be run.
-
 ## Contributing
 
 Contributions are welcome! Please feel free to submit a pull request.
@@ -177,32 +176,62 @@ Contributions are welcome! Please feel free to submit a pull request.
 When developing, use `pip install -e '.[dev]'` to install dev dependencies such
 as linter and formatter.
 
-- How to run tests:
+### How to run tests:
 
 ```sh
 pytest
 ```
 
-- How to run some of the tests, and re-run them whenever you change a file:
+### How to run some of the tests, and re-run them whenever you change a file:
 
 ```sh
 tp -i test ... # replace with path to tests/directories
 ```
 
-- How to run locally without XPK:
+
+### How to run HuggingFace transformer models
+Torchprime supports run with huggingface models by taking advantage of `tp run`.
+To use huggingface models, you can clone
+[huggingface/transformers](https://github.com/huggingface/transformers) under
+torchprime and name it as `local_transformers`. This allows you to pick any
+branch or make code modifications in transformers for experiment:
+```
+git clone https://github.com/huggingface/transformers.git local_transformers
+```
+If huggingface transformer doesn't exist, torchprime will automatically clone
+the repo and build the docker for experiment. To switch to huggingface models,
+add flag `--use-hf` to `tp run` command:
+```
+tp run --use-hf torchprime/hf_models/train.py
+```
+
+### How to run inside the docker container locally
+You can also run locally without XPK with docker. When running inside the docker
+container, it will use the same dependencies and build process as used in the
+XPK approach, improving the hermeticity and reliability.
+```
+tp docker-run torchprime/torch_xla_models/train.py
+```
+This will run the TorchPrime docker image locally. You can also add `--use-hf`
+to run HuggingFace model locally.
+```
+tp docker-run --use-hf torchprime/hf_models/train.py
+```
+
+### How to run locally without XPK:
 ```
 tp dbrun torchprime/torch_xla_models/train.py
 ```
 This will run the TorchPrime docker image locally. You can also add `--use-hf`
 to run HuggingFace model locally.
 
-- How to format:
+### How to format:
 
 ```sh
 ruff format
 ```
 
-- How to lint:
+### How to lint:
 
 ```sh
 ruff check [--fix]
--- a/README.md
+++ b/README.md
@@ -177,25 +177,32 @@ Contributions are welcome! Please feel free to submit a pull request.
 When developing, use `pip install -e '.[dev]'` to install dev dependencies such
 as linter and formatter.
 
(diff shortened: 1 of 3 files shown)


torchprime/launcher/cli.py
```diff
@@ -194,6 +194,45 @@ def create_and_activate_gcloud(gcloud_config_name, config: Config):
...
+def dbrun(args, use_hf: bool):
+  """
+  Runs the provided training command locally for quick testing.
+  """
+  config = read_config()
+
+  click.echo(get_project_dir().absolute())
+
+  # Build docker image.
+  build_arg = "USE_TRANSFORMERS=true" if use_hf else None
```
zpcore, line 213: I didn't reuse `def run(args, name: str | None, use_hf: bool):` since there is no need for interactive run in this case, and the code redundant part is not that big.

tengyifei:
I usually run a model locally like this:

```
python3 torchprime/torch_xla_models/train.py model=llama-3-8b mesh.fsdp=8 profile_step=3 max_steps=50
```

Is this insufficient for your use case?

zpcore:
> I usually run a model locally like this:
> 
> ```
> python3 torchprime/torch_xla_models/train.py model=llama-3-8b mesh.fsdp=8 profile_step=3 max_steps=50
> ```
> 
> Is this insufficient for your use case?

Yes, this should also work. The thing is that I saw many permission issues when run train.py directly. This just helps to run in a container.

README.md
```diff
@@ -144,25 +144,32 @@ Contributions are welcome! Please feel free to submit a pull request.
...
 
-How to run some of the tests, and re-run them whenever you change a file:
+- How to run some of the tests, and re-run them whenever you change a file:
 
 ```sh
 tp -i test ... # replace with path to tests/directories
 ```
 
-How to format:
+- How to run locally without XPK:
```
tengyifei, line 159: If you're adding more advanced ways of running models here, could you move "## Run huggingface transformer models" down here too, and make it a "###" H3 heading.

README.md
```diff
@@ -144,25 +144,32 @@ Contributions are welcome! Please feel free to submit a pull request.
 When developing, use `pip install -e '.[dev]'` to install dev dependencies such
 as linter and formatter.
 
-How to run tests:
+- How to run tests:
```
tengyifei, line 147: I think it's more readable if these are H3 heading ("###"):

```
### How to run tests
```

torchprime/launcher/cli.py
```diff
@@ -194,6 +194,57 @@ def create_and_activate_gcloud(gcloud_config_name, config: Config):
...
 
 
+@cli.command(
+  context_settings=dict(
+    ignore_unknown_options=True,
+  )
+)
+@click.argument("args", nargs=-1, type=click.UNPROCESSED)
+@click.option("--use-hf", is_flag=True, help="Use HuggingFace transformer")
+def dbrun(args, use_hf: bool):
```
tengyifei, line 204: It's unclear what "dbrun" stands for. Perhaps we could write

```
@cli.command(
  name="docker-run",
  context_settings=dict(
    ignore_unknown_options=True,
  )
)
```

Which will make the command "tp docker-run ..."

README.md
zpcore: Do you mean move `### Run huggingface transformer models` under `## Contributing`?

tengyifei, line 159: I think a clear description can be "How to run locally inside a docker container"

We can also point out a benefit of this approach, which is "When running inside the docker container, it will use the same dependencies and build process as used in the XPK approach, improving the hermeticity and reliability."

torchprime/launcher/cli.py
```diff
@@ -194,6 +194,57 @@ def create_and_activate_gcloud(gcloud_config_name, config: Config):
...
+  # Build docker image.
+  build_arg = "USE_TRANSFORMERS=true" if use_hf else None
+  docker_project = config.docker_project
+  if docker_project is None:
+    docker_project = config.project
+  docker_url = buildpush(
+    docker_project, torchprime_docker_tag="local_run", build_arg=build_arg
+  )
+  # Forward a bunch of important env vars.
+  env_forwarding = [
```
tengyifei, line 221: Could we factor this out into a `_get_env_forwarding()` helper? This way we only have one list to maintain.

torchprime/launcher/cli.py
```diff
@@ -194,6 +194,57 @@ def create_and_activate_gcloud(gcloud_config_name, config: Config):
...
+    *forward_env("XLA_IR_DEBUG"),  # torch_xla debugging flag
+    *forward_env("XLA_HLO_DEBUG"),  # torch_xla debugging flag
+    *forward_env("LIBTPU_INIT_ARGS"),  # XLA flags
+  ]
+  command = [
+    "python",
+  ] + list(args)
+  docker_command = [
+    "docker",
+    "run",
```
tengyifei, line 232: Do you think it's easier if we add a `-i` flag after `docker run`? I think that will run this docker image in an interactive context, meaning the outputs from the docker container is interactively shown in the terminal.

tengyifei (changes requested):

README.md
tengyifei: Yeah, I figured that might more of a contributor thing. Regular user trying out torchprime will just want to run torchprime.

torchprime/launcher/buildpush.py
```diff
@@ -61,7 +61,8 @@ def buildpush(
     _run(
       f"{sudo_cmd} docker tag {docker_tag} {docker_url}",
     )
-    _run(f"{sudo_cmd} docker push {docker_url}")
+    if torchprime_docker_tag != "local_run":
```
tengyifei, line 64: This introduces a magical constant. I think it's simpler if we add a "push=True" function argument, and have the other file call this with push=False

tengyifei (approved):

Issue #112, opened 2025-02-14
[torch_xla] Performance of Llama 3.1 405B matches Huggingface fork on 2 pods


tengyifei:
The MFU of https://github.com/pytorch-tpu/transformers/tree/flash_attention_405b on 2 pods is 18.46%

PR #117, opened 2025-02-20
Llama 3.1 405B 2 pod support

tengyifei:
# Llama 3.1 405B 2 pod support

This PR adds 2 pod support for Llama 3.1 405B. Because the best
performing Llama configuration does FSDP over DCN links, we need to
re-adjust the virtual device mesh to support FSDP over DCN. I've
extended the mesh to a pair of two dicts, `mesh` and `dcn_mesh`, each
of which may contain `data`, `fsdp`, etc.

To get better performance, we also need a very specific device ID
arrangement in the virtual device mesh. Specifically, we need each group
of 4 TPUs used in tensor parallelism to form a ring. On a 16x16 Trillium
pod, the device IDs will look like:

```
0, 1, 17, 16,
2, 3, 19, 18,
4, 5, 21, 20,
...
```

and so on. Note that the group `0, 1, 17, 16` will end up forming a ring
because the TPU with ID 16 is in practice physically located at the first
column and the second row. This replicates the MaxText feature at [1].

Tested:

```
export LIBTPU_INIT_ARGS='--xla_enable_async_all_gather=true --xla_tpu_enable_async_collective_fusion=true --xla_tpu_enable_async_collective_fusion_fuse_all_gather=true --xla_tpu_enable_async_collective_fusion_multiple_steps=true --xla_tpu_decompose_all_gather_einsum=true --xla_tpu_decompose_einsum_reduce_scatter=true --xla_tpu_scoped_vmem_limit_kib=98304 --xla_tpu_spmd_rng_bit_generator_unsafe=true --xla_tpu_overlap_compute_collective_tc=true --xla_tpu_use_enhanced_launch_barrier=true'

tp run torchprime/torch_xla_models/train.py model=llama-3.1-405b global_batch_size=128 dcn_mesh.fsdp=2 mesh.fsdp=64 mesh.tensor=4 dataset_config_name=wikitext-103-raw-v1 profile_step=15 profile_duration=200000 max_steps=50 logging_steps=10
```

The step time is 30.933s (see [2]), which is better than the 31.822s
in the Hugging Face fork of
https://github.com/pytorch-tpu/transformers/tree/flash_attention_405b.

Fixes https://github.com/AI-Hypercomputer/torchprime/issues/112.

[1]: https://github.com/AI-Hypercomputer/maxtext/pull/972/files

[2]: http://shortn/_Yy2QuIpkus


torchprime/torch_xla_models/configs/default.yaml
```diff
@@ -25,8 +25,25 @@ optimizer:
 lr_scheduler:
   type: linear
   warmup_steps: 0
+
+# The virtual device mesh shape to use within a TPU slice. This is also called
+# the "ICI mesh", since devices within a slice enjoy a faster network called
+# "Inter-Chip Interconnect".
 mesh:
```
bhavya01, line 32: nit: Should we just rename this to ici_mesh?

bhavya01 (approved):

torchprime/tests/test_custom_mesh.py
```diff
@@ -0,0 +1,251 @@
...
+    dcn_mesh_shape=(1, 2, 1, 1, 1),
+    num_devices=512,
+    num_slices=2,
+  )
+  assert (devices == get_64x4_reference_device_ids_2pod()).all()
+
+
+def get_64x4_reference_device_ids_1pod():
+  """Expected device arrangement for a 1 pod 64x4 mesh."""
+  return np.array(
```
pgmoka, line 48: What is this array based on?

tengyifei: This represents the most efficient ways to perform collectives in a 64x4 group. Specifically, the devices in each row (0, 1, 17, 16) are arranged in a ring on the TPU. A ring topology is more efficient than a line.

I realized I should document this (essentially what's in the PR description) into a `.md` file: https://github.com/AI-Hypercomputer/torchprime/issues/124

torchprime/torch_xla_models/configs/default.yaml
tengyifei: Renamed

PR #126, opened 2025-02-27
add llama3.1-8b config

zpcore:
Add the model config for llama 3.1 8B.

The config is referred from https://huggingface.co/meta-llama/Llama-3.1-8B/tree/main


bhavya01 (approved):
Nit: Can we add huggingface config for reference in future to compare easily?

zpcore:
> Nit: Can we add huggingface config for reference in future to compare easily?

Do you mean add the reference in the PR description? Thanks

zpcore:
> Nit: Can we add huggingface config for reference in future to compare easily?

I see what you mean, let me add the HF config in the same PR.

bhavya01:
I just meant in the PR description.

PR #138, opened 2025-03-01
Document model status

tengyifei:
The main change is the addition of a "Model Status" table in the README.md. Because it's good to show how to run a specific model if it's been optimized, I then added recipes for each model and linked them.

I'm also proposing these high level stages for supporting a model:

- **TODO**: We need to implement the model.
- **Implemented**: The model runs either a training or an inference step.
- **Optimized**: We found the best scaling configuration for the model on one or
  more hardware. One-off performance data is available.
- **Convergence**: We tested that the training loss converges to a reasonable
  value.
- **Production**: Not only is the model optimized and converges, its performance
  is also continuously monitored. This is a good state for using the model in
  production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,25 +120,31 @@ each model:
 - **Optimized**: We found the best scaling configuration for the model on one or
   more hardware. One-off performance data is available.
 - **Convergence**: We tested that the training loss converges to a reasonable
-  value.
+  value, or that the loss curve tracks an existing reference if exists.
 - **Production**: Not only is the model optimized and converges, its performance
   is also continuously monitored. This is a good state for using the model in
   production.
 
-If a model is at least implemented, you'll also find a training recipe linked
-from the table.
+All implemented models will at least have unit tests to verify basic numerical
+correctness, and the convergence verification stage serves as an additional
+correctness guarantee.
 
-| **Model**            | **torch_xla**                                                                                                                                                | **torchax**                                                                                         |
-| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


cloudchrischan:
Awesome!

* Can you create two table columns for "Correct" and "Optimized," and use checkmarks ✅ 
* Not in the scope of this PR, but would like to figure out what it will take to make torchax not an explicit choice, but rather something a user opts into via their desire to use specific JAX features. So in the (hopefully) near future, the user doesn't make an explicit decision about which framework they are using (beyond PyTorch, of course)

tengyifei:
Here's how it looks now:

screenshot-2025-03-03-10-47-12

There are multiple levels of correctness: the most basic is unit testing the model output, which we have for all implemented models. The next level is convergence. The final level is model performance benchmark such as MMLU and HellaSwag. Since the unit test correctness is very basic, I've took "correct" to mean convergence in this table.

I removed the torchax columns both to reduce bloat and also in the spirit of making JAX features opt-in in the future.

bhavya01:
A general comment I think that optimized is always on ongoing task. It's probably not helpful to have a column for that. Or probably better to just have an MFU and throughput column. WDYT?

README.md
```diff
@@ -183,6 +201,28 @@ ruff check [--fix]
...
+```
+
+If huggingface transformer doesn't exist, torchprime will automatically clone
+the repo and build the docker for experiment. To switch to huggingface models,
+add flag `--use-hf` to `tp run` command:
+
+```sh
+tp run --use-hf torchprime/hf_models/train.py
+```
+
```
bhavya01, line 225: Would it be cleaner to add this to `hf_models/README.md` and mention `torchprime/hf_models` in structure section? Just hoping to keep the main README.md show high level information

tengyifei: I like this idea. It's also in line with confusing newcomers less.

bhavya01 (approved):

Issue #66, opened 2025-01-30
Test torchprime on multiple TPU pods


bhavya01:
* Verify that minibatch sharding of data works correctly
*  Check profile
* Verify that loss curve looks good

tengyifei:
I think it's reasonable to fix this by adding a multi-slice E2E test now that we have E2E test support.

tengyifei:
Closing in favor of https://github.com/AI-Hypercomputer/torchprime/issues/169

PR #146, opened 2025-03-08
Add a multi-slice E2E test and update torch_xla pin to 20250313

tengyifei:
This can catch MegaScale related regressions.

Part of https://github.com/AI-Hypercomputer/torchprime/issues/66.

### DIFF
--- a/.github/workflows/cpu_test.yml
+++ b/.github/workflows/cpu_test.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11"]
     container:
-      image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_${{ matrix.python-version }}_tpuvm_cxx11_20250312
+      image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_${{ matrix.python-version }}_tpuvm_cxx11_20250313
     steps:
     - uses: actions/checkout@v4
     - name: Install torchax
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -17,6 +17,7 @@ jobs:
     outputs:
       llama-3-8b-name: ${{ steps.run-llama-3-8b.outputs.name }}
       llama-3-8b-2d-name: ${{ steps.run-llama-3-8b-2d.outputs.name }}
+      llama-3-8b-2-slice-name: ${{ steps.run-llama-3-8b-2-slice.outputs.name }}
       mixtral-8x7b-name: ${{ steps.run-mixtral-8x7b.outputs.name }}
       artifact-dir: ${{ steps.artifacts.outputs.artifact_dir }}
     steps:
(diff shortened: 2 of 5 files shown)


.github/workflows/cpu_test.yml
```diff
@@ -15,12 +15,9 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11"]
     container:
-      image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_${{ matrix.python-version }}_tpuvm_cxx11_20250307
+      image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_${{ matrix.python-version }}_tpuvm_cxx11_20250114
```
miladm, line 18: can we make the pin date explicitely called out as config?

torchprime/launcher/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:experimental
 # Use torch_xla Python 3.10 as the base image
-FROM us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.10_tpuvm_cxx11_20250307
+FROM us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.10_tpuvm_cxx11_20250114
```
miladm, line 3: hopefully we define this date in one place to prevent unintended mismatch?

torchprime/launcher/cli.py
```diff
@@ -273,7 +284,7 @@ def run(args, name: str | None, use_hf: bool, use_local_wheel: bool):
       "--tpu-type",
       config.tpu_type,
       "--num-slices",
-      str(config.num_slices),
+      str(num_slices),
```
miladm, line 287: how many slices do we test on? would be nice to do both 1 and 2 slices by deafult.

tengyifei: Yes, we used to test on 1 slice and now this PR will add both 1 slice and 2 slice testing.

.github/workflows/cpu_test.yml
tengyifei: https://github.com/AI-Hypercomputer/torchprime/issues/153

torchprime/launcher/cli.py
```diff
@@ -258,6 +266,9 @@ def run(args, name: str | None, use_hf: bool, use_local_wheel: bool):
     f"TORCHPRIME_JOBSET_NAME={workload_name}",
   ]
 
+  if num_slices is None:
+    num_slices = config.num_slices
```
zpcore, line 270: I think `config.num_slices` should have higher priority. `--num_slices` seems to be a hack only for E2E test.  Do we plan to expose `--num_slices`? Prefer hide the option from users to make it simple.

tengyifei: My intention is indeed to expose `--num-slices`. This way I don't need to run `tp use ...` again and wait 3 seconds to change the slice count. I could directly do it for one command and fallback to default otherwise. Do you think there's a downside I didn't realized?

zpcore: I think as long as we make num_slice in `tp use ...` and `tp run ...` clear, that should be good.  E.g., `tp run --num-slices` decides on the final number slices we run on.

zpcore (approved):
LGTM!

torchprime/launcher/cli.py
tengyifei: Let me add a documentation in both..

tengyifei: ty

.github/workflows/e2e_test.yml
```diff
@@ -102,6 +103,28 @@ jobs:
             profile_step=3 \
             max_steps=15
 
+      - name: Run Llama 3.0 8B (2 slice)
```
miladm, line 106: how can we avoid hard coded model names? can we have a yaml file with all candidate models that can subscribe to the multi-slice solution?

tengyifei: In principle all models could be tested with multi-slice but we only have 256 chips in `us-central2-b` at the moment (64 v6e-4 => 32 2-slice runs) and we'd need to account for parallel PRs and test runs etc. I filed https://github.com/AI-Hypercomputer/torchprime/issues/164 that should at least make it simpler.

PR #151, opened 2025-03-12
Introduce two FX graph partitioners: remat_all and remat_all_and_offload_these_inputs

tengyifei:
The main purpose is to be able to add activation checkpointing and host offloading of layer inputs to scan, which will be added later. Specifically, `torch_xla` scan takes a FX graph partitioner, so we can pass `remat_all` to implement full activation checkpointing, and pass `remat_all_and_offload_these_inputs` to implement "full activation checkpointing + offloading some input tensors to host".

And add unit tests that they work as intended.

### DIFF
--- a/torchprime/torch_xla_models/offloading.py
+++ b/torchprime/torch_xla_models/offloading.py
@@ -6,6 +6,7 @@ import torch.fx as fx
 from functorch.compile import aot_function, make_boxed_func  # type:ignore
 from torch.utils._pytree import tree_iter
 from torch.utils.checkpoint import CheckpointPolicy
+from torch_xla.experimental.stablehlo_custom_call import place_to_device, place_to_host
 
 from .remat_all import remat_all_partition_fn
 
@@ -56,7 +57,9 @@ def remat_all_and_offload_these_inputs(
   names_to_offload_set = set(names_to_offload)
 
   # Modify the module such that all `offload_name` tensors whose name match
-  # `names_to_offload_set` must be saved during the forward pass.
+  # `names_to_offload_set` must be saved during the forward pass. Then these
+  # nodes will show up in the output of the `fwd` graph as additional
+  # residuals. Later, we'll walk over the graph output to identify the nodes.
   for node in joint_module.graph.nodes:
     if (
       node.op == "call_function"
@@ -74,8 +77,8 @@ def remat_all_and_offload_these_inputs(
     fw_example_args = _make_arguments(fwd)
     bw_example_args = _make_arguments(bwd)
 
-  fw_name_in_output_indices = _get_name_in_output_indices(fwd)
-  bw_name_in_input_names = _get_name_in_input_names(
+  fw_name_in_output_indices = _get_offload_name_to_fw_output_indices(fwd)
+  bw_name_in_input_names = _get_offload_name_to_bw_input_names(
     bwd, num_fwd_outputs, fw_name_in_output_indices
   )
 
@@ -91,12 +94,18 @@ In the backward graph:
   """
 
   for name in names_to_offload_set:
-    assert name in fw_name_in_output_indices, _debug(
-      f"Did not find {name} in fw_name_in_output_indices: {fw_name_in_output_indices}."
-    )
-    assert name in bw_name_in_input_names, _debug(
-      f"Did not find {name} in bw_name_in_input_names: {bw_name_in_input_names}."
-    )
+    if name not in fw_name_in_output_indices:
+      raise ValueError(
+        _debug(
+          f"Did not find {name} in fw_name_in_output_indices: {fw_name_in_output_indices}."
+        )
+      )
+    if name not in bw_name_in_input_names:
+      raise ValueError(
+        _debug(
+          f"Did not find {name} in bw_name_in_input_names: {bw_name_in_input_names}."
+        )
+      )
 
   with torch.no_grad():
 
@@ -106,11 +115,7 @@ In the backward graph:
         [fw_name_in_output_indices[name] for name in names_to_offload_set]
       )
       return tuple(
-        torch.ops.xla.place_to_host(v)  # type: ignore
-        if i  # type:ignore
-        in indices_to_offload
-        else v
-        for i, v in enumerate(out)
+        place_to_host(v) if i in indices_to_offload else v for i, v in enumerate(out)
       )
 
     def backward(**kwargs):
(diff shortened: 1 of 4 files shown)


torchprime/torch_xla_models/tests/test_remat_all.py
```diff
@@ -0,0 +1,132 @@
...
+
+  # Forward graph does these ops
+  assert _count_function_calls(fw_graph, "mm") == 1
+  assert _count_function_calls(fw_graph, "sin") == 1
+  assert _count_function_calls(fw_graph, "exp") == 1
+
+  # Backward graph does these ops
+  assert _count_function_calls(bw_graph, "mm") == 3  # 1 forward + 2 backward
+  assert _count_function_calls(bw_graph, "sin") == 1  # 1 forward
+  assert _count_function_calls(bw_graph, "exp") == 1  # 1 backward (gradient of exp)
```
zpcore, line 41: why there is no forward of exp here?

torchprime/torch_xla_models/remat_all.py
```diff
@@ -0,0 +1,80 @@
...
+        # graphs at once. We only have one graph so this can simply be 0.
+        node.meta["ac_graph_id"] = 0
+
+    return min_cut_rematerialization_partition(
+      joint_module, _joint_inputs, num_fwd_outputs=num_fwd_outputs
+    )
+
+
+@contextmanager
+def _remat_all_config():
```
zpcore, line 37: Is this necessary even though we have applied `CheckpointPolicy.MUST_RECOMPUTE`?

torchprime/torch_xla_models/tests/test_remat_all.py
tengyifei: Eagle eyes! I think this is the result of chain rule. So the backward function gets dLoss/dc. Then the remat part computes b. Then to get dLoss/db, we run dLoss/db = dLoss/dc * dc/db and dc/db works out to be dexp(b)/db = exp(b). So there's only one exp during this process. 


Actually, having typed this out, I think no matter what final operation we put in f (be it exp or sin or cos or whatever), the dLoss/db will never depend on the value of c. Therefore we never have to recompute c again.

torchprime/torch_xla_models/offloading.py
```diff
@@ -0,0 +1,224 @@
...
+from functorch.compile import aot_function, make_boxed_func  # type:ignore
+from torch.utils._pytree import tree_iter
+from torch.utils.checkpoint import CheckpointPolicy
+from torch_xla.experimental.stablehlo_custom_call import place_to_device, place_to_host
+
+from .remat_all import remat_all_partition_fn
+
+
+@torch.library.custom_op("xla::offload_name", mutates_args=())
+def offload_name(t: torch.Tensor, name: str) -> torch.Tensor:
```
zpcore, line 15: Could you point me to the code how the `name` argument become the `_qualname` attribute of the fx graph node?

zpcore (commented):
The logic LGTM! Left few questions to help me better understand the code.

torchprime/torch_xla_models/offloading.py
```diff
@@ -0,0 +1,224 @@
...
+    return None
+  return torch.empty_like(t)
+
+
+@offload_name.register_autograd
+def _offload_name_backward(ctx, grad):
+  return grad, None
+
+
+def remat_all_and_offload_these_inputs(
```
zpcore, line 41: Since we can choose to offload tensors based on the assigned `name`, what stops us from offloading activation inside the decoder layer?

torchprime/torch_xla_models/tests/test_remat_all.py
tengyifei: Explained in comment

zpcore: Yes, exactly! We never need to recompute the last output whatever the function is during backpropagation.

torchprime/torch_xla_models/offloading.py
tengyifei: Actually `name` becomes the second argument of the node, and `offload_name._qualname` is supposed to be equal to `torch.ops.xla.offload_name` or something. Therefore, later we check `node.target.name() == offload_name._qualname`. That check effectively means "is this node a function call and this node is calling the offload_name" function.

I'll document this next to the `_qualname` thing

torchprime/torch_xla_models/offloading.py
tengyifei: Good point. My first prototype can't do it because I didn't figure out how to control the rematerialization. Now we can. I'll add a test to test offloading a middle tensor.

Now, why we couldn't do this before? It turns out that's because my earlier prototype [1] rematted _everything_, including the `offload_name` call itself. As in the partitioned forward and backward graph looks like this:

```
def forward(input, primal_1, primal_2, ...):
  offloaded = offload_name(input, "foo")
  mm = aten.mm(offloaded, primal_1)
  ...
  return input, primal_1, ...

def backward(input, primal_1, primal_2, ...):
  # even offload_name is run again
  offloaded = offload_name(input, "foo")
  mm = aten.mm(offloaded, primal_1)
  ...
```

Notice how there are _two_ offload_name calls. In order to find the `input` tensor to offload, I walk backwards from `offload_name` in both graphs to find the tensor which is the input to `offload_name`.

When I tried to use the same code with Llama 405B it didn't work for some reason. That prompted me to adjust how the remat works, using the `CheckpointPolicy` instead of min-cut configs. After that, the graph looks like this:

```
def forward(input, primal_1, primal_2, ...):
  offloaded = offload_name(input, "foo")
  mm = aten.mm(offloaded, primal_1)
  ...
  return offloaded, primal_1, ...

def backward(offloaded, primal_1, primal_2, ...):
  # no more `offload_name` here
  mm = aten.mm(offloaded, primal_1)
  ...
```

Notice because the policy for `offload_name` is set to `CheckpointPolicy.MUST_SAVE`, the graph partitioner always saves the `offloaded` tensor in the context. Then during the backward pass, it directly uses that tensor instead of using `input`, because `input` is discarded while `offloaded` is saved. Now I just have to look for an `offload_name` call in the output of the `forward`, and look for the corresponding argument index in the input of the `backward`, then wrap them with the place_to_host and place_to_device calls.

This is precisely what enables us now to offload some intermediate tensor because whichever tensor got decorated with `offload_name` and which we requested to offload will get tagged with the `CheckpointPolicy.MUST_SAVE` policy. This functionality didn't exist in the earlier prototype.

[1]: https://github.com/pytorch-tpu/transformers/blob/flash_attention_405b-dev/src/transformers/models/llama/graph_transforms/offloading.py

tengyifei:
PTAL

zpcore (approved):
Thanks for all those detailed explanations!

PR #156, opened 2025-03-17
Wire up CLI for scan and host offloading

tengyifei:
In the previous PR (https://github.com/AI-Hypercomputer/torchprime/pull/157) we added the ability to replace a sequence of layers with scan and in an earlier PR (https://github.com/AI-Hypercomputer/torchprime/pull/151) we added the ability to offload tensors by transforming a graph. This PR changes the CLI to be able to select either for-loop, scan, or scan+host offloading.

In order to mix-and-match `sharding` configs with scan/offloading configs without creating a combinatorial explosion, I had to add a new config group called `remat` and add scan and offloading there. Since `activation_checkpointing_layers` and the optimization barrier is also part of remat, I moved it there too. This means the only left config in the `scaling` group is `sharding`, so I've hoisted `sharding` to be top level. This moves a bunch of files but it's mostly `s/scaling/sharding`.


torchprime/torch_xla_models/train.py
```diff
@@ -171,25 +172,65 @@ def _get_train_dataloader(self):
...
-
-    logger.info(f"Enabling activation checkpointing on {classes}")
+    layers_to_scan = self.config.model.remat.get("scan_layers", None)
+    offload_tensors = self.config.model.remat.get("offload_tensors", [])
+
+    # Checking preconditions and logging.
+    if remat_classes:
+      logger.info(f"Enabling activation checkpointing on {remat_classes}")
+    if layers_to_scan:
+      assert isinstance(layers_to_scan, str)
```
zpcore, line 186: Based on the logic, it only supports one module to be scanned. Is this true?

zpcore (approved):
LGTM! Left one question.

torchprime/torch_xla_models/configs/default.yaml
```diff
@@ -47,3 +47,16 @@ dcn_mesh:
...
+# in the `model/remat` directory.
+model:
+  remat:
+    # If specified, compile a module of type `HomogeneousSequential` located at the
+    # given path in the module tree using `torch_xla.experimental.scan_layers`.
+    scan_layers: null
+
+    # If specified, offload these tensors to host RAM during the forward pass and
+    # move them back during the backward pass.
+    offload_tensors: []
```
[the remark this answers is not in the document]
tengyifei: ah, yes

torchprime/torch_xla_models/train.py
tengyifei: > Then there can only be one layer to be offloaded.

There can only be one _sequence_ of homogenous layers whose individual layers have offloading.

For example, if `layers_to_scan` is `model.layers`, then that will match the `HomogenousSequential` object that wraps e.g. 126 decoder layers in case of Llama 3.1 405B. Even though we can only add offloading to one sequence of layers, this ends up cover 126 individual decoder layers.

For LLM decoders this works nicely since most of the tricks center around the decoder layer. But I imagine for other fancier models this would not be sufficient.

PR #157, opened 2025-03-17
Make models amenable to scan

tengyifei:
We replace the `for` loop in both Llama and Mixtral with an equivalent `HomogenousSequential` layer, which can be either run a for loop or use `torch_xla`'s scan operator. This is a clean-ish way to turn scan on/off without cluttering the modeling code.

I also adjusted Mixtral slightly so that we can even run `scan` in Mixtral with its static MoE implementation. In order to integrate with scan, we need to refactor the Mixtral decoder for loop into a format where results from the previous iteration feed into the next iteration. Scanning over GMM on the other hand won't work until GMM forward/backward is wrapped in a custom op similar to https://github.com/pytorch/xla/pull/8654.

Cleanup the README that got jumbled in https://github.com/AI-Hypercomputer/torchprime/pull/111 while I'm here.

Test: added unit test. Next PR will change the trainer to apply scan.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,7 @@ Finally, each model may also provide a GPU "original" version that illustrates
 and attributes where this model code came from, if any. This also helps to
 show case what changes we have done to make it performant on TPU. The original
 version is not expected to be run.
+
 ## Contributing
 
 Contributions are welcome! Please feel free to submit a pull request.
@@ -176,62 +177,25 @@ Contributions are welcome! Please feel free to submit a pull request.
 When developing, use `pip install -e '.[dev]'` to install dev dependencies such
 as linter and formatter.
 
-### How to run tests:
+### How to run tests
 
 ```sh
 pytest
 ```
 
-### How to run some of the tests, and re-run them whenever you change a file:
+### How to run some of the tests, and re-run them whenever you change a file
 
 ```sh
 tp -i test ... # replace with path to tests/directories
 ```
 
-
-### How to run HuggingFace transformer models
-Torchprime supports run with huggingface models by taking advantage of `tp run`.
-To use huggingface models, you can clone
-[huggingface/transformers](https://github.com/huggingface/transformers) under
-torchprime and name it as `local_transformers`. This allows you to pick any
-branch or make code modifications in transformers for experiment:
-```
-git clone https://github.com/huggingface/transformers.git local_transformers
-```
-If huggingface transformer doesn't exist, torchprime will automatically clone
-the repo and build the docker for experiment. To switch to huggingface models,
-add flag `--use-hf` to `tp run` command:
-```
-tp run --use-hf torchprime/hf_models/train.py
-```
-
-### How to run inside the docker container locally
-You can also run locally without XPK with docker. When running inside the docker
-container, it will use the same dependencies and build process as used in the
-XPK approach, improving the hermeticity and reliability.
-```
-tp docker-run torchprime/torch_xla_models/train.py
-```
-This will run the TorchPrime docker image locally. You can also add `--use-hf`
-to run HuggingFace model locally.
-```
-tp docker-run --use-hf torchprime/hf_models/train.py
-```
-
-### How to run locally without XPK:
-```
-tp dbrun torchprime/torch_xla_models/train.py
-```
-This will run the TorchPrime docker image locally. You can also add `--use-hf`
-to run HuggingFace model locally.
-
-### How to format:
+### How to format
 
 ```sh
 ruff format
 ```
 
-### How to lint:
+### How to lint
 
 ```sh
 ruff check [--fix]
@@ -239,6 +203,23 @@ ruff check [--fix]
 
 You can install a Ruff VSCode plugin to check errors and format files from
 the editor.
+
+### How to run inside the docker container locally
+
+You can also run locally without XPK with docker. When running inside the docker
+container, it will use the same dependencies and build process as used in the
+XPK approach, improving the hermeticity and reliability.
+
+```sh
+tp docker-run torchprime/torch_xla_models/train.py
+```
+
+This will run the TorchPrime docker image locally. You can also add `--use-hf`
+to run HuggingFace model locally.
+
+```sh
+tp docker-run --use-hf torchprime/hf_models/train.py
+```
 
 ## Run distributed training with local torch/torch_xla wheel
 
--- a/torchprime/layers/sequential.py
+++ b/torchprime/layers/sequential.py
@@ -1,0 +1,41 @@
+from typing import Any
+
+import torch.nn as nn
+
+PyTree = Any
+
+
+class HomogeneousSequential(nn.Sequential):
+  """
+  HomogenousSequential is a sequential container that requires all child modules
+  to be of the same type and have matching input/output shapes. In turn, it may be
+  compiled with the `scan` higher order operator to save compile time.
+  """
+
+  repeated_layer: type
+  """The type of the layer being looped over."""
+
+  def __init__(self, *args: nn.Module) -> None:
+    super().__init__(*args)
+    types = set(type(module) for module in args)
+    assert len(types) == 1, f"All modules must be of the same type. Got {types}"
+    self.repeated_layer = types.pop()
+
+  def forward(self, *input, **broadcasted_inputs: PyTree):
+    """
+    Much like `torch.nn.Sequential`, this takes `input` and forwards it to the
+    first module it contains. It then "chains" outputs to inputs sequentially for
+    each subsequent module, finally returning the output of the last module.
+    Different from `torch.nn.Sequential`, you may specify `broadcasted_inputs` via
+    keyword arguments. The same keyword arguments will be passed to every layer
+    without changes (i.e. "broadcasted").
+    """
+    for module in self:
+      input = module(*splat(input), **broadcasted_inputs)
+    return input
+
+
+def splat(input):
+  if not isinstance(input, list | tuple):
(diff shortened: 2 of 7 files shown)


torchprime/torch_xla_models/mixtral/model.py
```diff
@@ -978,12 +990,12 @@ def _init_weights(self, module):
...
   @xp.trace_me("MixtralModel")
   def forward(
-    self, input_ids: torch.LongTensor = None, attention_mask: torch.Tensor | None = None
+    self, input_ids: torch.LongTensor, attention_mask: torch.Tensor | None = None
   ) -> tuple:
     batch_size, seq_length = input_ids.shape
 
     device = input_ids.device
-    position_ids = torch.arange(seq_length, dtype=torch.long, device=device)
+    position_ids = torch.arange(seq_length, dtype=torch.long, device=device).float()
```
bhavya01, line 998: Can you explain why are we changing these to float? I am hoping it doesn't have any effect on the other parts of the profile

torchprime/torch_xla_models/mixtral/model.py
```diff
@@ -32,6 +32,7 @@
 from torch.nn import init
 
 from torchprime.torch_xla_models.loss import cross_entropy_loss
+from torchprime.torch_xla_models.scan_layers import HomogeneousSequential
```
bhavya01, line 35: Can we change this to

```
from torchprime.layers.sequential import HomogeneousSequential
```

torchprime/torch_xla_models/scan_layers.py
```diff
@@ -0,0 +1,55 @@
...
+from torchprime.layers.sequential import HomogeneousSequential, PyTree, splat
+
+
+class HomogeneousSequentialScan(HomogeneousSequential):
+  def __init__(self, *args, partition_fn=default_partition):
+    super().__init__(*args)
+    self.partition_fn = partition_fn
+
+  def forward(self, *input, **broadcasted_inputs: PyTree):
+    layers = [BroadcastArguments(m) for m in self.children()]
```
bhavya01, line 15: Can you elaborate on what self.children() will refer to in this case? I can't figure it out by reading the code.

torchprime/torch_xla_models/mixtral/model.py
tengyifei: Good catch, this is due to a limitation of `scan`, which currently requires all inputs to be differentiable. Let me add a reference to https://github.com/pytorch/xla/issues/8783. We can go back to using `long` when https://github.com/pytorch/xla/issues/8783 is fixed.

torchprime/torch_xla_models/mixtral/model.py
tengyifei: IDE auto import got too happy

torchprime/torch_xla_models/scan_layers.py
tengyifei: Added a comment. It just returns the sequence of decoder layers. So this way we wrap each in a `BroadcastArguments`, which is a trick to make the decoder layers work with `scan`. In the `BroadcastArguments` we extend each decoder layer's output to also output the broadcasted inputs (position IDs, etc). This plumbs those values across scan iterations so the same position IDs are available to all layers.

README.md
```diff
@@ -169,77 +169,58 @@ Finally, each model may also provide a GPU "original" version that illustrates
...
-To use huggingface models, you can clone
-[huggingface/transformers](https://github.com/huggingface/transformers) under
-torchprime and name it as `local_transformers`. This allows you to pick any
-branch or make code modifications in transformers for experiment:
-```
-git clone https://github.com/huggingface/transformers.git local_transformers
-```
-If huggingface transformer doesn't exist, torchprime will automatically clone
-the repo and build the docker for experiment. To switch to huggingface models,
-add flag `--use-hf` to `tp run` command:
```
zpcore, line 203: Is this removed intentionally? I saw you also mentioned `--use-hf` later.

tengyifei: I relocated this part to https://github.com/AI-Hypercomputer/torchprime/tree/main/torchprime/hf_models because PM doesn't want us to publicize HF in torchprime

torchprime/torch_xla_models/mixtral/model.py
zpcore: My understanding is that as long as the `hidden_states` doesn't contribute to the `input_ids`, `attention_mask` in the forward pass, enabling the grad for `input_ids`, `attention_mask` should be fine based on the chain rule.

torchprime/torch_xla_models/tests/test_llama.py
```diff
@@ -32,7 +32,7 @@ def get_llama_3_8b() -> LlamaFixture:
     intermediate_size=16,
     vocab_size=vocab_size,
   )
-  config.attention_kernel = None
+  config.flash_attention = False
```
zpcore, line 35: I remember updated the config from `config.flash_attention` into `config.attention_kernel` in the splash attention support PR. Can we use `config.attention_kernel`?

torchprime/torch_xla_models/mixtral/model.py
tengyifei: Actually PyTorch will discard the gradients when we compute them in the autograd Function for `scan`. If PyTorch sees that `Scan.backwards` returns a non-None value for an input carry, but the input (e.g. attention_mask) does not require grad, then PyTorch will disregard that value. That means we can safely "compute" those gradients and it will still be a no-op. But I wanted to solve https://github.com/pytorch/xla/issues/8783 by properly returning None for those gradients to be safe.

torchprime/torch_xla_models/tests/test_mixtral.py
```diff
@@ -10,98 +12,133 @@
...
+  vocab_size = 128
+  config = AutoConfig.from_pretrained(
+    "mistralai/Mixtral-8x7B-v0.1",
+    num_hidden_layers=1,
+    num_attention_heads=8,
+    hidden_size=8,
+    intermediate_size=16,
+    vocab_size=vocab_size,
+  )
+  config.flash_attention = False
```
zpcore, line 34: Maybe update to `config.attention_kerne`l?

torchprime/torch_xla_models/tests/test_llama.py
tengyifei: I think the `config` is a HuggingFace config object (i.e. from the transformers repo), so that one still takes `flash_attention`. WDYT?

torchprime/torch_xla_models/tests/test_mixtral.py
tengyifei: Ditto

torchprime/torch_xla_models/mixtral/model.py
```diff
@@ -915,9 +923,10 @@ def __init__(self, config: DictConfig, layer_idx: int):
   def forward(
     self,
     hidden_states: torch.Tensor,
+    cumulative_loss: torch.Tensor,
```
zpcore, line 926: Why we update the forward with `cumulative_loss`?

tengyifei: I did this to be able to use `scan` in Mixtral. In order to use scan, we require the previous layer's output to be fed into the next layer. For llama this works great (previous layer outputs hidden -> send hidden to the next layer). But for Mixtral, the hidden state is being passed to the next layer, but the load balance loss is being accumulated (see [1]). This doesn't fit the mold of passing everything from one layer to the next.

So I solve it by slightly refactoring the logic to have one layer pass the load balance loss to the next layer. Each layer adds on to the loss, and the last layer returns the total load balance loss. This refactor will work with either for loop or scan.

[1]: https://github.com/AI-Hypercomputer/torchprime/pull/157/files/51d4568620b20e804224c91fcb3ad2c09efbff4e#diff-f9b00c4a4f6b8795083b0ec66711ad66fb93c2b1dfd04c2ec46f9a6153ce6a14L1002

torchprime/torch_xla_models/scan_layers.py
bhavya01: I see. It is coming from nn.Sequential that is the parent class of HomogeneousSequential 👍

bhavya01 (approved):

torchprime/layers/sequential.py
```diff
@@ -0,0 +1,41 @@
...
+  repeated_layer: type
+  """The type of the layer being looped over."""
+
+  def __init__(self, *args: nn.Module) -> None:
+    super().__init__(*args)
+    types = set(type(module) for module in args)
+    assert len(types) == 1, f"All modules must be of the same type. Got {types}"
+    self.repeated_layer = types.pop()
+
+  def forward(self, *input, **broadcasted_inputs: PyTree):
```
zpcore, line 24: Seems like it is the user responsibility convert `broadcasted_inputs` to the tensor type that can require_grad for scan.

tengyifei: Yes. Once https://github.com/pytorch/xla/issues/8783 is fixed, the user won't have to do any manual conversion.

tengyifei:
@zpcore i saw you added a number of comments but didn't press "Request Changes" or "Approve" -- let me know if you would like to request changes or approve.

torchprime/torch_xla_models/tests/test_llama.py
zpcore: Got it, this is only for HF model.

zpcore (approved):
LGTM!

PR #158, opened 2025-03-18
Update mixtral implementation to dropping

bhavya01:
Step time with the current implementation: 8.931s with the command:
```
 tp run torchprime/torch_xla_models/train.py global_batch_size=2048 max_steps=50 block_size=4096 profile_step=10 ici_mesh.fsdp=256 dataset_config_name=wikitext-103-raw-v1
```

Run logs: https://cloudlogging.app.goo.gl/e56NzVU29jv5D9Ef9

The best huggingface step time that I got was 8.691s. I think that the difference could be due to different libtpu versions and I don't think it's worth spending additional time to optimize further.

### DIFF
--- a/torchprime/torch_xla_models/mixtral/model.py
+++ b/torchprime/torch_xla_models/mixtral/model.py
@@ -30,6 +30,7 @@ import torch_xla.distributed.spmd as xs
 from omegaconf import DictConfig
 from torch import nn
 from torch.nn import init
+from torch_xla.distributed.spmd.xla_sharding import MarkShardingFunction
 
 from torchprime.layers.sequential import HomogeneousSequential
 from torchprime.torch_xla_models.loss import cross_entropy_loss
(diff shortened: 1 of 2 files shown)


torchprime/torch_xla_models/mixtral/model.py
```diff
@@ -358,13 +359,12 @@ def forward(self, dispatch_input):
     mesh = xs.get_global_mesh()
     assert mesh is not None
     layer_w1 = torch.einsum("ebcm,emh->ebch", dispatch_input, self.w1)
-    xs.mark_sharding(layer_w1, mesh, ("expert", ("data", "fsdp"), None, None))
+    MarkShardingFunction.apply(layer_w1, mesh, ("expert", ("data", "fsdp"), None, None))
```
tengyifei, line 362: If `MarkShardingFunction` is no longer in-place as of https://github.com/pytorch/xla/pull/8821, wouldn't we need to write it like

```
layer_w1 = MarkShardingFunction.apply(layer_w1, mesh, ("expert", ("data", "fsdp"), None, None))
```

If we don't use the return value of `MarkShardingFunction`, I suspect `layer_w1` won't be sharded at all. The E2E test probably passed because we're not checking memory usage and it happened to fit on a v6e-4. Do you think that could be the case?

tengyifei (changes requested):

tengyifei (approved):
LGTM. Need to fix lint via `ruff check` / `ruff format`.

PR #201, opened 2025-04-09
Publish optimized Llama 3.1 405B recipe

### LINKED ISSUE #160: Llama 3.1 405B scan+offloading OOMs
Directly turning on scan+offloading in Llama 3.1 405B causes the training to OOM on the backward pass. The culprit is computing the gradient of the loss w.r.t. the weight of the `o_proj` layer. This gradient is only 1D sharded and not 2D sharded. That then infects a bunch of other ops to be also 1D sharded and causes the OOM.

https://github.com/AI-Hypercomputer/torchprime/commit/350f20e4abef199fa4b5ec68afea599dbe6ad5b8 contains an ugly hack to avoid the OOM.

We need to figure out how to avoid this to be able to achieve 32% MFU in Llama 3.1 405B via scan+offloading.

tengyifei:
Do the last bits of updates for Llama 3.1 405B to enable host offloading.

Update the recipes and the MFU numbers.

Also tweak the README while I'm at it.

Fixes #160.

Unfortunately the 2 pod run still hangs if we enable profiling, so I manually calculated the step time by reading the logs: [1].

My 1 pod run got terminated by the cluster but I was also able to recover the step time by reading the logs: [2].

[1]: http://shortn/_2HOCrq22rp
[2]: http://shortn/_46LE1f5x90


tengyifei:
I'll need to retry/fix the Apr 9 `torch_xla` nightly build and then update the pin here to `20250409`.

tengyifei:
> I'll need to retry/fix the Apr 9 `torch_xla` nightly build and then update the pin here to `20250409` to pickup the latest libtpu.

OK, this is ready now.

zpcore (approved):
LGTM!

PR #203, opened 2025-04-11
use docker-run locally without tp config

zpcore:
1. Now we can run `tp docker-run` locally without `tp use` prerequisite;
2. Remove jax dependency from Dockerfile because torchax is installed with torch_xla together now.

### DIFF
--- a/torchprime/launcher/buildpush.py
+++ b/torchprime/launcher/buildpush.py
@@ -5,6 +5,7 @@ import getpass
 import grp
 import os
 import random
+import re
 import string
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


torchprime/launcher/buildpush.py
```diff
@@ -41,9 +42,13 @@ def buildpush(
   # Determine Docker URL
   default_url = f"gcr.io/{torchprime_project_id}/torchprime-{user}:{docker_tag}"
   docker_url = torchprime_docker_url if torchprime_docker_url else default_url
+  docker_url = re.sub(r"/+", "/", docker_url).lower()
```
tengyifei, line 45: What does this do? Can we add a comment or see if it's necessary?

torchprime/launcher/cli.py
```diff
@@ -215,15 +215,22 @@ def docker_run(args, use_hf: bool):
   """
   Runs the provided training command locally for quick testing.
   """
-  config = read_config()
+  run_with_tp_config = True
+  try:
```
tengyifei, line 219: If `docker-run` does not require a config, I think we should not attempt to read the config. Catching exceptions would gloss over actual errors.

It looks like the code is reading the config to get the `docker_project`, and it needs the `docker_project` to call `buildpush`. I think a cleaner way to phrase this is to make the `docker_project` argument optional in `buildpush`, and only require a non-None value if `push_docker` is `True`. Then we don't have to read the config.

tengyifei (changes requested):

torchprime/launcher/cli.py
tengyifei: If the docker building step still requires a URL, I think we could give docker a placeholder URL instead of requiring a project from the config.

torchprime/launcher/buildpush.py
zpcore: The docker_url doesn't accept `//` and upper case. So I convert consective `/` into a single slash. Added the comment.

tengyifei (approved):

PR #204, opened 2025-04-11
add llama3.1 70B config

zpcore:
Add the llama3.1 70B config, will update the performance number next.

Config is based on https://huggingface.co/meta-llama/Llama-3.1-70B/blob/main/config.json.


zpcore:
I think we probably don't need a specific E2E test just this model.

PR #207, opened 2025-04-12
Bump tensorflow to 2.19

tengyifei:
Looks like installing 2.18 will forcefully downgrade jax to 0.53, which nullifies the recent libtpu pin update.

TEST: llama 405b 1 pod profile: http://shortn/_BlyRalgivF, logs: http://shortn/_lCU1qeMBjz


zpcore:
Oh, that's probably the root cause of why JAX get uninstalled. Thanks!

PR #208, opened 2025-04-13
Check step duration in E2E test

### LINKED ISSUE #170: Check step duration in E2E tests
We should obtain some historical reference and check that the step duration does not fall too high above that range.

tengyifei:
Fixes #170

Now CI and the E2E test will fail if the performance suddenly got too good or too bad.

At least in torchprime, we'll fail the E2E test even if the performance got too good (e.g. after a pin update). This is a reminder for the engineer to update these performance bounds so that we can closely track the latest SOTA.

I pulled a number of recent runs and did a T-distribution modeling to set the upper and lower bounds at 99.9% confidence interval: [1]. I only used data that's after a recent performance improvement.

[1]: https://docs.google.com/spreadsheets/d/1VS_0tkmmUwT22F2eeqqIOuExiBwIIbqn58MXw3RMe7M/edit?usp=sharing


README.md
```diff
@@ -120,6 +120,8 @@ Besides forwarding your command line arguments, `tp run` will add:
 
 - `profile_dir=[...]`: path to a [profile][torch_xla_profile] directory
   accessible by the workload
+- `output_dir=[...]` path to a directory where the workload may store output
```
zpcore, line 123: This must be gcs path right? Can we also allow local path if we run model locally?

tengyifei: Yes, we can always pass a local path in local debugging. It's just that `tp run` will pre-fill a GCS path for both profile_dir and output_dir now.

zpcore: I see, so both are allowed. Thanks!

zpcore (approved):
LGTM!

tengyifei:
> What is we have compile time metric ready in the future, is step time still the only metric to decision whether new PR can be merged or not?

I think it's reasonable to check all sorts of metrics. In fact, I purposefully named the step "check metrics" and created the starting point for a "training_metrics.json", where we could add more metrics in the future!

PR #213, opened 2025-04-23
Compatibility for 2.7 release

zpcore:


tengyifei (commented):
I think the old import works fine. Does it stop working in 2.8.0 for some reason? The versioning check is extra complexity.

zpcore:
Unfortunately, `jax_env_context` is not under path `torch_xla.experimental.custom_kernel` anymore in the PyTorch/XLA head.

tengyifei:
ugh. you're right

tengyifei (approved):

tengyifei:
Later on, it would be good to add an E2E test verifying certain recipes with the 2.7 docker. Otherwise, we'd easily regress and random things may start breaking as changes get landed.

tengyifei:
Maybe we could add an install sh script that downloads the right nightly version wheel corresponding to the pinned docker and installs them. To make the local installation process easier.

PR #214, opened 2025-04-24
update llama performance

zpcore:


torchprime/torch_xla_models/README.md
```diff
@@ -38,45 +38,63 @@ launch it on XPK.
 
 ### Llama 3.0 8B on v6e-256
 
-Recipe for global batch size 256, sequence length 8192.
-Expected step duration: 1.625s. MFU: 33.53%.
+Recipe for global batch size 1024, sequence length 8192. We need splash attention kernel.
```
yaoshiang, line 41: Do you believe the previous numbers correct? what enabled such a large change in the batch size, the splash attention kernel? Have we flexed the batch size?

zpcore: I think previously we didn't try the largest batch size. In addition, when we change from flash attention to splash attention, backward time can be significant reduced, which results in the MFU improvement.

zpcore: In fact, this can be further improvement when we disable the host offloading with `model/remat=llama-scan`

tengyifei: Was there some customer wanting the 256 gbs numbers? If so, we'd need to preserve that configuration.

torchprime/torch_xla_models/README.md
```diff
@@ -38,45 +38,64 @@ launch it on XPK.
...
+    ici_mesh.fsdp=256 \
+    model/remat=llama-scan \
+    model.attention_kernel=splash_attention
 ```
 
-Recipe for global batch size 512, sequence length 8192.
-Expected step duration: 2.991s. MFU: 36.43%.
+### Llama 3.1 8B on v6e-256
+
+Recipe for global batch size 1024, sequence length 8192. We need splash attention kernel.
```
tengyifei, line 60: nit: replace "We" with "This recipe uses the"

tengyifei (approved):

torchprime/torch_xla_models/README.md
zpcore: I am basically trying on the largest batch/device we can support to show the performance. I don't remember any customer is particular interested in 256 gbs.

tengyifei: Ack

torchprime/torch_xla_models/README.md
zpcore: I don't remember any customers want the specific gbs. @yaoshiang , do you happen to know any?

yaoshiang: no, and, I can't imagine a customer wanting a target BS per accelerator. The answer to batch size is almost always, fill up the memory. Sometimes you may want a target global batch size for optimization purposes, in that case you can use grad accumulation etc. But for pure MFU... usually big but less than OOM is what you want. 

(the nuance is that if 2x the batch size results in 50% the speed, then you don't need to push up the batch size anymore)

torchprime/torch_xla_models/README.md
zpcore: Oh, I forgot to update the "This recipe uses the". Will stress in the next PR together.

PR #215, opened 2025-04-24
Code snippet for running with r2.7 release

zpcore:
This is the code snippet to make torchprime runnable with PyTorch/XLA r2.7 release. No plan to merge just for reference.

### DIFF
--- a/torchprime/launcher/Dockerfile
+++ b/torchprime/launcher/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:experimental
 # Use torch_xla Python 3.10 as the base image
-FROM us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.10_tpuvm_20250410
+FROM us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:r2.7.0_3.10_tpuvm
 
 ARG USE_TRANSFORMERS=false
 ARG USE_LOCAL_WHEEL=false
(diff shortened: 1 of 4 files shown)


zpcore:
Maybe In the future, we can release torchprime to align with the PyTorch/XLA release schedule to make sure we have working version with PTXLA.

tengyifei:
IMO a simpler solution could be providing an installation script that automatically installs the right nightly wheels that matches the version used by E2E tests.

yaoshiang:
I actually think we should release PT/XLA frequently - so we should release a 2.7.1 in [2/4] weeks, etc. Along with my "pinned dep on torch" proposal, users wouldn't have to think about which version of torch to match up with... installing torch_xla just install the right version of torch. 

I don't think there is consensus on that, however.

In the meantime, as we add more scenarios to torch prime, I think each will need their own installation instructions with pins to specific nightlies we know work. The generic latest "nightly" would arguably be not as optimal as a specific nightly, since nightlies are by definition not stable.

tengyifei:
> so we should release a 2.7.1 in [2/4] weeks, etc. Along with my "pinned dep on torch" proposal

That's a good idea. In fact, I think we don't even have to be restricted to minor versions. E.g. if PyTorch releases `2.7.1`, we could do:
- `2.7.1.0501`
- `2.7.1.0515`
- etc
from HEAD periodically

And with the guarantee that they all work (ABI compatible and tested) with 2.7 and also the PyTorch HEAD.

PR #216, opened 2025-04-24
Add Transact model and model runner to torchprime

qihqi:
This commit proposes the following convention:

1. Models should reside in `models` folder.
2. Models should be registered in models.registry with str name and callable to construct that model. a. For naming conventions, prefer huggingface model id if available if not prefer the github org/repo_name if forked.
3. Every model in a subfolder (with potential, model specific configs or layers).
4. Shareble layers should go to models/layers
5. The model.py file in a model subfolder should contain the actual model
6. The model_original.py file should be a "almost verbatim" copy of the model from its source, this should be used for unit-testing correctness. By "almost verbatim" meaning the math should remain the same, but OK to change imports etc to make it run.
7. A model should has a method `get_sample_inputs` that can produce inputs of different batch size.

A series of scripts, will be able to use the above registry and run the models, the first one is introduced here to run the model's forward. More scripts will be added to run train loop, sanity checks, or micro benchmarks etc.

Moving models from torch_xla_models and torchax_models to follow the convention will come later.

### DIFF
--- a/e2e_testing/run_model_forward.py
+++ b/e2e_testing/run_model_forward.py
@@ -10,36 +10,37 @@ def run_model_torchax(model, batch_size, number_of_runs, eager):
   import jax
   import torchax as tx
   import torchax.interop
+
   torchax.enable_globally()
-  model = model.to('jax')
+  model = model.to("jax")
 
   wait_val = tx.interop.torch_view(jax.block_until_ready)
 
   if not eager:
     model = tx.interop.JittableModule(model)
 
-
   for i in range(number_of_runs):
-    print('Iteration ', i)
+    print("Iteration ", i)
     inputs = model.get_sample_inputs(batch_size)
-    args, kwargs = pytree.tree_map_only(torch.Tensor, lambda t: t.to('jax'), inputs)
+    args, kwargs = pytree.tree_map_only(torch.Tensor, lambda t: t.to("jax"), inputs)
     args, kwargs = wait_val((args, kwargs))
     start = time.perf_counter()
     res = model.forward(*args, **kwargs)
     wait_val(res)
     print(res)
     end = time.perf_counter()
-    print(f'Iteration {i} took: {end - start}s')
+    print(f"Iteration {i} took: {end - start}s")
 
 
 def run_model_xla(model, batch_size, number_of_runs):
   import torch_xla
-  model = model.to('xla')
+
+  model = model.to("xla")
 
   for i in range(number_of_runs):
-    print('Iteration ', i)
+    print("Iteration ", i)
     inputs = model.get_sample_inputs(batch_size)
-    args, kwargs = pytree.tree_map_only(torch.Tensor, lambda t: t.to('xla'), inputs)
+    args, kwargs = pytree.tree_map_only(torch.Tensor, lambda t: t.to("xla"), inputs)
     torch_xla.sync(wait=True)
     start = time.perf_counter()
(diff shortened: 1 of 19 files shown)


yaoshiang (approved):
I won't hold up this PR but I think we will soon need to consider a deeper refactor of the organization of this repo to be more decoupled, more of a monorepo, less of a single tool with a single requirements.txt and approach. In fact, to do competitive benchmarking, we may need very different requirements.txt at times (e.g. a pytorch-cuda requirements.txt to benchmark some model on GB200). Some models may need PTXLA-nightly, some may work out of the box on the latest stable release (e.g. 2.6.1).

e2e_testing/run_model_forward.py
```diff
@@ -0,0 +1,82 @@
+import time
```
tengyifei, line 1: Is this exercised by CI? The Llama and Mixtral models are tested on (CPU) CI with a small configuration via `pytest`. See e.g. https://github.com/AI-Hypercomputer/torchprime/blob/main/torchprime/torch_xla_models/tests/test_llama.py

I think in order to test this in CI we'd need a file with the `test_` prefix.

tengyifei (changes requested):

e2e_testing/run_model_forward.py
qihqi: the `run_model_forward` script is meant to run as a CLI script. I added a CI entry to test it directly.

tengyifei: Hmm, not sure why the E2E test doesn't pick up this file.

FWIW, we've been using E2E test to mean "exercise distributed training workflows", so actually the VM the E2E test runs on only submits jobs.

Would it be possible to write this `run_model_forward.py` as a parameterized pytest test? I think if you put it in `torchprime/tests` and give it a name like `test_run_model.py`, then 

https://github.com/AI-Hypercomputer/torchprime/blob/main/.github/workflows/cpu_test.yml#L50

will pick it up automatically.

qihqi: done. PTAL

torchprime/models/transact/transact_config.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass
+
+
+@dataclass
```
tengyifei, line 4: Could we add a run_model test that exercises a tiny version of the transact model? I looked around and only found a run_model test for a toy module but not sure if I missed something.

e2e_testing/run_model_forward.py
```diff
@@ -0,0 +1,39 @@
...
+
+  print("Model run times:")
+  for i, t in enumerate(times):
+    print(f"Iteration {i}: {t}s")
+
+  print("Model run successful")
+
+
+if __name__ == "__main__":
+  import fire
```
tengyifei, line 37: We need to add `fire` to the pyproject toml dependencies. This would've been caught if there was a test exercising this file (e.g. maybe in .github/workflows/cpu_tests.yaml)

tengyifei (approved):
LGTM modulo comments. 

Would you like to file a GitHub issue tracking the proposal made in this PR?

e2e_testing/run_model_forward.py
qihqi: done.

torchprime/models/transact/transact_config.py
qihqi: done.

PR #217, opened 2025-04-27
Use upstream splash attention

### LINKED ISSUE #153: Make the torch_xla docker pin date a config
Rather than hardcoding it in the docker file and github actions, it should be possible to write it down in one shared location and configure/override it locally as well.

tengyifei:
Also this is the N-th time I manually searched for the location of the
docker URL. I'll factor that out into a config.

Fixes #153.


zpcore (approved):
Thanks for cleaning up the code!

PR #225, opened 2025-05-03
Update Llama 3.1 405B recipe

tengyifei:
After updating to the May 1 nightly, the profiling crash is finally fixed.

Log of a successful run: http://shortn/_LtdUSg95GX


tengyifei:
Hmm, looks like the `Llama 3.1 8B 2 slice` step time has improved since the last pin update. Now it consistently finishes in 4 seconds or so. I'll make a separate PR to tighten the E2E step time.

vlasenkoalexey (approved):

PR #232, opened 2025-05-08
Document the current metrics

tengyifei:
Explain how the step time is computed currently in torchprime.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,31 @@
+# How is training performance measured
+
+Training performance is a multi-faceted topic. `torchprime` measures the
+following metrics, and saves them at `${output_dir}/train_metrics.json` at the
+end of a training run.
+
+### `train_runtime`
+
+The total time of the training run (including compilation). In the JSON metrics,
+it is a floating point representing the number of seconds of this training run.
+
+### `step_execution_time`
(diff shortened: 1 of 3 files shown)


docs/metrics.md
```diff
@@ -0,0 +1,31 @@
...
+`torchprime` relies on the profiler to obtain the most accurate accelerator
+timing. At the end of a training run, if a profile was taken, the trainer will
+compute the time delta between the start of every training step on the
+accelerator, and the median value of those time delta will be used as the
+`step_execution_time`.
+
+To obtain the step execution time, set a non-negative value in the `profile_step`
+CLI argument to the trainer. Typically, you should use `profile_step` of at least
+2 to skip over the first two compilation steps, which introduces large idle gaps
+on the accelerator.
```
vlasenkoalexey, line 25: @tengyifei as far as I understand this is not quite inline with the expectation with the rest of CMCS as we had in todays disucssion. Do you plan to amend this in the future, or maybe add new metrics?

tengyifei: Yes, I do expect this would have to be amended. I just document what we do now so I could reference it in docs etc.

vlasenkoalexey (approved):

Issue #36, opened 2025-01-14
[torch_xla] Training loss is NaN under the AdamW optimizer


tengyifei:
See https://github.com/AI-Hypercomputer/torchprime/pull/34#pullrequestreview-2550889555.

If we use the AdamW optimizer, the loss is NaN at step 3 (after 2 steps).

In order to have a comparison with MaxText and also for general experimentation, it makes sense to support both Adafactor and AdamW optimizers.

tengyifei:
For later reference: I did a bit of debugging while trying this for Llama 405B:

- Added a function to check if optimizer state is finite: https://github.com/AI-Hypercomputer/torchprime/blob/a4d58b706caddd7328e449d6049370a2379c3aec/torchprime/torch_xla_models/train.py#L388
- Switched to AdamW like so: https://github.com/AI-Hypercomputer/torchprime/blob/yifeit/test/torchprime/torch_xla_models/train.py#L108
- The optimizer state blows up at step 3:

```
python3 torchprime/torch_xla_models/train.py model=llama-3-8b mesh.fsdp=8 profile_step=3 max_steps=50 logging_steps=1

...

Error executing job with overrides: ['model=llama-3-8b', 'mesh.fsdp=8', 'profile_step=3', 'max_steps=50', 'logging_steps=1']
Traceback (most recent call last):
  File "/home/yifeit/torch/torchprime/torchprime/torch_xla_models/train.py", line 385, in main
    trainer.train_loop()
  File "/home/yifeit/torch/torchprime/torchprime/torch_xla_models/train.py", line 236, in train_loop
    check_adafactor_state(self.optimizer)
  File "/home/yifeit/torch/torchprime/torchprime/torch_xla_models/train.py", line 404, in check_adafactor_state
    raise ValueError(
ValueError: Non-finite value detected in state for parameter Parameter containing:
tensor([[-0.0025,  0.0057,  0.0378,  ...,  0.0310,  0.0154, -0.0041],
        [-0.0057, -0.0376, -0.0242,  ...,  0.0239,  0.0170, -0.0245],
        [ 0.0243,  0.0266, -0.0293,  ...,  0.0164,  0.0294, -0.0273],
        ...,
        [ 0.0112, -0.0036,  0.0051,  ...,  0.0142,  0.0014,  0.0013],
        [-0.0376, -0.0215, -0.0098,  ..., -0.0170,  0.0270,  0.0095],
[...]

PR #237, opened 2025-05-10
Update README.md

yaoshiang:
Removes all references to MFU. 

I think that it is very important that our repo always represents an accurate state of the repo and product. These numbers are applied science - the numbers must be meaningful to consumers. These MFU numbers used a memory efficient optimizer that is not proven to converge as well as AdamW.

### DIFF
--- a/torchprime/torch_xla_models/README.md
+++ b/torchprime/torch_xla_models/README.md
@@ -41,7 +41,6 @@ launch it on XPK.
 ### Llama 3.0 8B on v6e-256
 
 Recipe for global batch size 1024, sequence length 8192.
-Expected step duration: 4.741s. MFU: 45.98%.
 
 ```sh
 export LIBTPU_INIT_ARGS='--xla_tpu_scoped_vmem_limit_kib=98304 --xla_tpu_use_minor_sharding_for_major_trivial_input=true --xla_tpu_relayout_group_size_threshold_for_reduce_scatter=1 --xla_tpu_assign_all_reduce_scatter_layout=true --xla_tpu_enable_data_parallel_all_reduce_opt=true --xla_tpu_data_parallel_opt_different_sized_ops=true --xla_tpu_enable_async_collective_fusion=true --xla_tpu_enable_async_collective_fusion_fuse_all_gather=true --xla_tpu_enable_async_collective_fusion_multiple_steps=true  --xla_tpu_overlap_compute_collective_tc=true  --xla_enable_async_all_gather=true --xla_tpu_enable_async_collective_fusion_fuse_all_reduce=false  --xla_tpu_enable_sparse_core_collective_offload_all_reduce=true --xla_tpu_use_tc_device_shape_on_sc=true  --xla_sc_enable_instruction_fusion=false  --xla_sc_disjoint_spmem=false  --xla_sc_disable_megacore_partitioning=true  --2a886c8_chip_config_name=megachip_tccontrol'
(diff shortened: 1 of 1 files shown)


tengyifei (commented):
I get the sentiment but have questions:

- Right now https://github.com/AI-Hypercomputer/torchprime?tab=readme-ov-file#supported-models lists that these models are optimized and we publish performance. If we remove the published performance, what should we do in the model status table?

- Granted I don't have evidence, but the extra memory utilization from AdamW compared to AdaFactor is just one extra param. The perf will probably be quite similar once we fix https://github.com/AI-Hypercomputer/torchprime/issues/36.


Maybe instead we call out that the optimizer is Adafactor and we're working on AdamW numbers?

yaoshiang:
> I get the sentiment but have questions:
> 
> * Right now https://github.com/AI-Hypercomputer/torchprime?tab=readme-ov-file#supported-models lists that these models are optimized and we publish performance. If we remove the published performance, what should we do in the model status table?
> * Granted I don't have evidence, but the extra memory utilization from AdamW compared to AdaFactor is just one extra param. The perf will probably be quite similar once we fix [[torch_xla] Training loss is NaN under the AdamW optimizer #36](https://github.com/AI-Hypercomputer/torchprime/issues/36).
> 
> Maybe instead we call out that the optimizer is Adafactor and we're working on AdamW numbers?

I think there is some consensus that the ordering of "implemented, optimized, converges" should switch to "implemented, converges, optimized", and, that a model's not really complete until it has all three (plus documentation). Until then, it is work in progress, and best not to publish. That's a stronger statement than this PR, however, - we should discuss as a team.

Re: this PR, I don't think it makes sense to describe Adafactor for Llama because one of the main reference points was the Llama3 paper itself, which listed IIRC 38%-43% MFU with Adam. The memory usage of Adam is material, and the batch size is probable to go down.

I'd say I have a 25% confidence that that the MFU numbers will change materially under Adam - put another way, I only have a 75% confidence level that the number will hold up under what I think is the correct experimental setup... so that's not high enough confidence for me to publish. 

Totally understand if you aren't convinced... let's discuss as a team next week.

tengyifei:
I think there's also a product factor i.e. if we already told someone about certain MFU numbers, would they come here and find that those are now gone vs moved to an appendix with a disclaimer that it's optimized with Adafactor for now. cc @cloudchrischan

tengyifei:
I'll defer to @vlasenkoalexey or @cloudchrischan to approve the PR

cloudchrischan:
Aligned with:

1. Removing reference MFU
2. Reverting all "Optimized" status indicators to a TODO state

cloudchrischan:
@yaoshiang Can you revert the README status indicators to TODO in this PR?

cloudchrischan (approved):

cloudchrischan:
Approved, @yaoshiang confirmed updating the table in another PR

qihqi (approved):

vlasenkoalexey:
Agree that MFU probably should not be stated on the README page, especially if we have some doubts that numbers would hold if we change optimizer, which we most likely have to do. Also Yifei did non-trivial amount of work to achieve those results and it is not clear when we'll be able to achieve same results on AdamW. 

Therefore I'd suggest to move those numbers to another place, https://github.com/AI-Hypercomputer/torchprime?tab=readme-ov-file#supported-models could be a good one with asterisk that MFU is achieved using adafactor optimizer which is more memory efficient.

yaoshiang:
> Agree that MFU probably should not be stated on the README page, especially if we have some doubts that numbers would hold if we change optimizer, which we most likely have to do. Also Yifei did non-trivial amount of work to achieve those results and it is not clear when we'll be able to achieve same results on AdamW.
> 
> Therefore I'd suggest to move those numbers to another place, https://github.com/AI-Hypercomputer/torchprime?tab=readme-ov-file#supported-models could be a good one with asterisk that MFU is achieved using adafactor optimizer which is more memory efficient.

I've added the numbers to a metrics tracking CSV including Adafactor as a column.

PR #238, opened 2025-05-10
Add SFT 

jialei777:


torchprime/torch_xla_models/llama/model.py
```diff
@@ -58,6 +58,86 @@ def load_sharded_safetensors_to_state_dict(model_dir: str) -> dict:
 
     return state_dict
 
+def save_sharded_safetensors_by_layer(
+    state_dict: dict, 
+    save_dir: str,
+    layer_keys: list = None  # List of layer prefixes to group by
```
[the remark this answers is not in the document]
yaoshiang: should we add an Optional or `| None` here? https://peps.python.org/pep-0604/#specification

yaoshiang (changes requested):
I'd suggest adding unit tests to verify that the weights are saved properly and reloadable. Loading HF pretrained weights, sampling some weights and/or sampling some outputs, then saving   the checkpoint and attempting to reload, and then verifying the same weight values and output. 

A few other style guide comments from the Google Python Style Guide and PEP8.

torchprime/torch_xla_models/configs/default.yaml
```diff
@@ -9,17 +9,17 @@ defaults:
 
 dataset_name: gsm8k
 dataset_config_name: main
-global_batch_size: 4
+global_batch_size: 16
```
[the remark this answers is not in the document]
vlasenkoalexey: +1 let's not change this

torchprime/data/dataset.py
```diff
@@ -19,12 +19,33 @@ def make_huggingface_dataset(
...
-  column_names = list(data.features)
-  data = data.map(
-    lambda samples: tokenizer(samples["text"]),
-    batched=True,
-    remove_columns=column_names,
-  )
+  def preprocess_batch(batch):
+    inputs, labels, attention_masks = [], [], []
+
+    for q, a in zip(batch["question"], batch["answer"]):
```
vlasenkoalexey, line 25: is this format specific to gsm8k dataset? 
I guess we should have a generic design that is not limited to one dataset

torchprime/torch_xla_models/llama/model.py
```diff
@@ -33,6 +33,110 @@
...
+        prefixes = set()
+        for key in state_dict.keys():
+            parts = key.split('.')
+            if len(parts) >= 3 and parts[1] == 'layers' and parts[2].isdigit():
+                prefixes.add(f"{parts[0]}.{parts[1]}.{parts[2]}")
+        
+        layer_keys = sorted(list(prefixes))
+        
+        # Add embeddings and output layer groups
+        layer_keys = ['model.embed', 'model.norm'] + layer_keys + ['lm_head']
```
vlasenkoalexey, line 92: this should come from sharding config

torchprime/torch_xla_models/llama/model.py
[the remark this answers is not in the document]
vlasenkoalexey: FYI we agreed that @yaoshiang will work on checkpoint saving/loading

torchprime/torch_xla_models/llama/model.py
```diff
@@ -394,3 +498,96 @@ def forward(
       return logits, None
     loss = cross_entropy_loss(logits, labels=labels, vocab_size=self.config.vocab_size)
     return logits, loss
+
+  @classmethod
+  def from_pretrained(cls, config):
```
vlasenkoalexey, line 503: this should not be limited to Llama

vlasenkoalexey (commented):
thanks for putting it together, I don't think we can merge it in current state.
As discussed, could you write a proposal for introduction of SFT into torchprime based on what MaxText has. And we'll take it from there

torchprime/data/dataset.py
```diff
@@ -19,12 +19,33 @@ def make_huggingface_dataset(
...
+      prompt = f"Question: {q}\n"
+      full_text = prompt + f"Answer: {a}\n\n\n"
```
vlad-karp, line 27: prompt templates worth moving into config

torchprime/torch_xla_models/llama/model.py
vlad-karp: is this to mimic HuggingFace classes? We can probably do it in a more consistent way in the long-term

PR #243, opened 2025-05-19
Replacing torch.polar math by sin/cos in rope layer

vlasenkoalexey:
The should be sufficient to make llama4 training to work


vlasenkoalexey:
Filed https://github.com/pytorch/xla/issues/9215 to support torch.polar and related ops in PyTorch/XLA

qihqi (approved):

PR #246, opened 2025-05-21
Jackoh/upload benchmark db

jack8558:
##  Description 
This PR adds feature to upload benchmark result to bigquery. `thunk.py` is modified to call `upload_metrics_to_bq` after the training script is done.

- To use, add `--upload-metrics` flag in `tp use`
- Can also configure bigquery dataset in `tp use` with `--bq-project`, `--bq-dataset`, and `--bq-table`
- `--comments` in`tp run` for `logs_comments` column in the table

Test Bqtable: http://shortn/_5IEJjY8qIJ


torchprime/launcher/benchmark_db_util.py
```diff
@@ -0,0 +1,102 @@
...
+
+def get_model_name(args: list[str]) -> str:
+  """Extracts the model name from script arguments."""
+  # Iterate through the script arguments to find model name
+  for arg in args:
+    if arg.startswith("model="):
+      return arg.split("=", 1)[1]
+
+  # TODO: Load default model from a config file or a more robust mechanism
+  default_model = "llama-3-8b"
```
tengyifei, line 44: Why not save the model name into `metrics.json`? We could extend the `Metrics` class any way we like.

torchprime/launcher/benchmark_db_util.py
```diff
@@ -0,0 +1,102 @@
...
+      tpu_type: The type of TPU used (e.g., "v4-8").
+      kwargs: Rest of the schema that is included in the dictionary.
+  """
+  current_time_str = datetime.now().strftime("%Y%m%d%H%M%S")
+  unique_id_suffix = str(uuid.uuid4()).split("-")[0]  # Short UUID
+  run_id = f"{jobset_name}-{current_time_str}-{unique_id_suffix}"
+
+  model = get_model_name(script_args)
+
+  # TODO: Refactor as there is duplicate logic at compute_mfu in mfu.py
```
tengyifei, line 72: Could we factor out a helper function?

torchprime/launcher/benchmark_db_util.py
```diff
@@ -0,0 +1,102 @@
...
+  elif tpu_type.startswith("v5e"):
+    hardware_num_chips = core_count
+    tflops_per_chip = 197
+  elif tpu_type.startswith("v6e"):
+    hardware_num_chips = core_count
+    tflops_per_chip = 918
+  else:
+    raise ValueError(f"Unsupported accelerator type {tpu_type}")
+
+  print(
```
tengyifei, line 89: nit: we use `click.echo` which is guaranteed to flush for you

torchprime/launcher/cli.py
```diff
@@ -94,6 +99,31 @@ def cli(ctx, interactive):
...
+@click.option(
+  "--bq-dataset",
+  required=False,
+  default="benchmark_dataset_test",
+  help="A bigqeury dataset to upload metrics.",
+)
+@click.option(
+  "--bq-table",
+  required=False,
+  default="jackoh-test-tp-benchmark3",
```
tengyifei, line 124: Could we switch to a shared bigquery table in the `tpu-pytorch` project? Maybe you can create one for everybody?

torchprime/launcher/thunk.py
```diff
@@ -66,4 +68,23 @@
...
 env = os.environ.copy()
-os.execve(sys.executable, args, env)
+process = subprocess.run(args, env=env)
+
+# Upload result to database
+upload_metrics = os.getenv("TORCHPRIME_UPLOAD_METRICS")
+
+if upload_metrics.lower() == "true" and slice_id == "0" and worker_id == "0":
+  try:
+    print(
```
tengyifei, line 78: nit: `click.echo`

tengyifei (changes requested):

torchprime/launcher/cli.py
jack8558: I created a new table name "benchmark_experiment" for default as of now. Let me know if you have other preference.

torchprime/torch_xla_models/configs/model/llama-3-8b.yaml
```diff
@@ -22,3 +22,4 @@ attention_bias: false
 # choose attention_kernel from: [flash_attention, splash_attention, null]
 attention_kernel: flash_attention
 rope_theta: 500000.0
+torch_dtype: bfloat16
```
tengyifei, line 25: Is this field used anywhere?

torchprime/launcher/benchmark_db_util.py
jack8558: Great idea, as we might want to save model configuration in benchmark (e.g optimizer, sequence_length ), we can add them to metrics later as well.

torchprime/launcher/benchmark_db_util.py
```diff
@@ -0,0 +1,79 @@
...
+  Returns:
+      A dictionary containing model_id, metrics_step_time (in seconds),
+      and metrics_mfu, or None if the metrics file is not found.
+  """
+  metric_file_path = Path(
+    os.path.join(
+      base_artifact_path,
+      jobset_name_for_outputs,
+      "outputs",
+      "0-0",
```
vlasenkoalexey, line 30: is it always 0-0 ? 
any idea where this value is coming from?

torchprime/launcher/benchmark_db_util.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def prepare_benchmark_summary(
+  process_returncode: int,
+  jobset_name: str,
+  tpu_type: str,
+  **kwargs,
+) -> dict:
+  """
+  Constructs a summary dictionary of the benchmark run.
```
vlasenkoalexey, line 54: nit: add returns to the docstring

torchprime/launcher/benchmark_db_util.py
```diff
@@ -0,0 +1,79 @@
...
+  run_id = f"{jobset_name}-{current_time_str}-{unique_id_suffix}"
+
+  hardware_num_chips, tflops_per_chip = get_num_chips_and_tflops_per_chip(tpu_type)
+
+  click.echo(
+    f"Tpu type: {tpu_type}, hardware_num_chips: {hardware_num_chips}, tflops_per_chip: {tflops_per_chip}"
+  )
+
+  return {
+    "run_id": run_id,
```
vlasenkoalexey, line 73: is schema defined anywhere we can reference?

torchprime/torch_xla_models/configs/model/llama-3-8b.yaml
vlasenkoalexey: fyi: I don't think that setting this value has any impact on model precision in torchprime, probably you should use https://docs.pytorch.org/docs/stable/generated/torch.get_default_dtype.html#torch-get-default-dtype instead to check on precision

torchprime/launcher/thunk.py
```diff
@@ -66,4 +70,21 @@
...
 env = os.environ.copy()
-os.execve(sys.executable, args, env)
+process = subprocess.run(args, env=env)
+
+# Upload result to database
+upload_metrics = os.getenv("TORCHPRIME_UPLOAD_METRICS")
+
+if upload_metrics.lower() == "true" and slice_id == "0" and worker_id == "0":
+  try:
+    click.echo(
```
vlasenkoalexey, line 80: why not logger.info here and below?
any benefits of using click instead?

torchprime/launcher/upload_metrics_to_bq.py
```diff
@@ -0,0 +1,87 @@
...
+  Uploads the prepared benchmark summary to BigQuery.
+  This is the low-level client interaction.
+  Args:
+      summary_data: A dictionary containing the benchmark summary,
+                    typically from _prepare_benchmark_summary.
+      bq_project: The BigQuery project ID.
+      bq_dataset: The BigQuery dataset ID.
+      bq_table: The BigQuery table ID.
+  """
+  click.echo("Attempting to upload benchmark results to BigQuery...")
```
vlasenkoalexey, line 28: same question

vlasenkoalexey (changes requested):

torchprime/launcher/benchmark_db_util.py
jack8558: It is coming from thunk.py when we define output_dir. What if basically meanns is slice 0 and worker 0: https://github.com/AI-Hypercomputer/torchprime/blob/31c450e82c6273f50f9815351f6fbebb42903f58/torchprime/launcher/thunk.py#L56

So my understanding is that when we run torchprime, it will run on each worker. So for example, the "thunk.py train.py {args}" that is created in run method in cli.py will be running on each worker for all slices.
https://github.com/AI-Hypercomputer/torchprime/blob/31c450e82c6273f50f9815351f6fbebb42903f58/torchprime/launcher/cli.py#L320

And I think this means that metrics file will be exising for each worker. (If there is 1 slice and 4 worker, we get metrics file in each worker). If you see the line 78 and thunk.py, I am only uploading the benchmark result for slice 0 and worker 0. I realized if I don't do that, I was uploading rows of data for each worker. IIUC, each step time should be same (or very similar) so extracting values from one worker should be sufficient.

@tengyifei Please correct me if anything is wrong.

torchprime/launcher/thunk.py
jack8558: It was originally print statement but I changed to click.echo because of automatic flush. I am not sure about comparing it to logger.info actually. @tengyifei Hi Yifei, any thoughts on this?

torchprime/launcher/benchmark_db_util.py
tengyifei: It's fine to align with the MaxText choices, imo.

We could also create torchprime-specific filters or spreadsheets based on what's in workload_benchamrk_v2_schema to get simpler views!

torchprime/launcher/thunk.py
tengyifei: I don't have a strong opinion. It's just that all the launcher code so far has been using `click` while the trainer has been using `logger.info`, which adds a bunch of metadata like time and severity.

It could be fine to do a separate pass to move everything to `logger`.

torchprime/torch_xla_models/configs/model/llama-3-8b.yaml
tengyifei: Seems a good idea to get the default dtype from PyTorch.

There's an issue of transmitting that data from the trainer to `thunk.py`. I think we could do this:

- Write the model dtype in `metrics.json`
- Read it from `thunk.py` and upload it

torchprime/launcher/upload_metrics_to_bq.py
```diff
@@ -0,0 +1,87 @@
...
+  bq_project: str,
+  bq_dataset: str,
+  bq_table: str,
+):
+  """
+  Uploads the prepared benchmark summary to BigQuery.
+  This is the low-level client interaction.
+  Args:
+      summary_data: A dictionary containing the benchmark summary,
+                    typically from _prepare_benchmark_summary.
```
tengyifei, line 23: nit: `prepare_benchmark_summary` not `_prepare_benchmark_summary`, I think

torchprime/launcher/benchmark_db_util.py
jack8558: Thank you for the clarification Yifei.

I added the comment what 0-0 means. @vlasenkoalexey Let me know if it doesn't make sense, feel free to resolve this comment.

torchprime/torch_xla_models/configs/model/llama-3-8b.yaml
jack8558: Makese sense, since compute_mfu is called from train.py, I don't think we need to write the model dtype in metrics.json. I passed the dtype through model config after getting it from `get-default-dtype` and removed all the torch_dtype field from yaml files.

tengyifei: sgtm

vlasenkoalexey: in such case, could you please remove these config changes

torchprime/launcher/benchmark_db_util.py
vlasenkoalexey: I'd just put a comment in the code to make it clear, thanks

vlasenkoalexey:
Don't see any major issues, please wait for @tengyifei to sign off before merging. 
Thanks

vlasenkoalexey (approved):

torchprime/torch_xla_models/configs/model/llama-3-8b.yaml
tengyifei, line 25: Can we remove these `torch_dtype` now that this is no longer used?

torchprime/torch_xla_models/train.py
```diff
@@ -317,11 +323,27 @@ def step_closure(epoch, step, loss, trace_start_time, trace_end_time):
...
-    # Analyze the step duration from the latest profile
     if self.config.profile_step >= 0:
+      # Analyze the step duration from the latest profile
       step_duration = step_duration_from_latest_profile(self.config.profile_dir)
       metrics_logger.log_step_execution_time(step_duration)
 
+      # Compute the MFU.
+      tpu_name = os.environ.get("TORCHPRIME_TPU_TYPE", None)
+      self.config.model.torch_dtype = (
+        torch.get_default_dtype()
```
tengyifei, line 334: Unfortunately this won't be accurate; we'll need to inspect the dtype from the model weights. The reason is that line 381 (https://github.com/AI-Hypercomputer/torchprime/pull/246/files#diff-6e7ee464035cac83b0c1171dbea2511919602ed800005c2aa355aef72d04c54cR381) is used to set the dtype, and that decorator only temporarily sets the dtype.

Here's one quick way to get the dtype from the model:

```python
def get_dtype(module):
  dtypes = {param.dtype for param in module.parameters()}
  if len(dtypes) != 1:
    raise ValueError(f"Inconsistent dtypes found: {dtypes}")
  return dtypes.pop()
```

tengyifei (changes requested):
Could you modify the E2E test to exercise DB uploading? Basically update https://github.com/AI-Hypercomputer/torchprime/blob/cb1918711d7316445fdcb5fb6c7aaa9920053299/.github/actions/e2e-setup/action.yml#L64 to upload to a E2E specific bq table which you may also create from Pantheon.

This way, we'll be able to prevent bit-rotting of the upload logic.

In fact this will make fetching and analyzing the E2E step times much easier too!

tengyifei (approved):

PR #264, opened 2025-05-30
Clean up train.py: move logic to trainer, optimization, and sharding folder

jialei777:
- Simply moving trainer logic to a separate file `base_trainer.py`, no new logic added
- Move remat logic into separate folders
- Add unit test for the trainer.


torchprime/torch_xla_models/trainers/basic.py
```diff
@@ -0,0 +1,353 @@
...
+      batch_size=batch_size,
+      sampler=sampler,
+      drop_last=True,
+    )
+    loader = pl.MpDeviceLoader(
+      dataloader, self.device, input_sharding=self.input_sharding_spec
+    )
+    return loader
+
+  def _add_checkpoint_offload_scan_model(self, model: nn.Module):
```
tengyifei, line 170: Do you expect to duplicate these functionality in the upcoming SFT trainer as well? Just trying to understand the intended structure.

We probably need sharding/scan/optimization-barrier/etc in SFT too.

torchprime/torch_xla_models/trainer/basic.py
```diff
@@ -0,0 +1,233 @@
+import logging
```
yaoshiang, line 1: Please add a module docstring to describe the goal of this module

yaoshiang (changes requested):

torchprime/torch_xla_models/optimization/checkpointing.py
```diff
@@ -0,0 +1,88 @@
+import logging
```
yaoshiang, line 1: I'd recommend a different module name as checkpointing has two disjoint meanings

torchprime/torch_xla_models/optimization/checkpointing.py
```diff
@@ -0,0 +1,88 @@
...
+    remat_all.remat_all_partition_fn
+    if not offload_tensors
+    else partial(
+      offloading.remat_all_and_offload_these_inputs, names_to_offload=offload_tensors
+    )
+  )
+  return scan_layers.compile(model, layers_to_scan, partition_fn=partition_fn)
+
+
+def add_optimization_barriers(model, config):
```
yaoshiang, line 63: Please add typehints and docstrings to all functions

torchprime/torch_xla_models/optimization/checkpointing.py
```diff
@@ -0,0 +1,88 @@
...
+from transformers.trainer_pt_utils import get_module_class_from_name
+
+from torchprime.layers.sequential import HomogeneousSequential
+from torchprime.sharding.shard_model import wrap_module
+from torchprime.torch_xla_models import offloading, remat_all, scan_layers
+
+logger = logging.getLogger(__name__)
+
+
+def add_activation_checkpointing_and_scan(model, config):
```
vlasenkoalexey, line 15: add docstring and type annotations for public methods, that's the general Python guideline we should follow

torchprime/torch_xla_models/optimization/checkpointing.py
vlasenkoalexey: +1

vlasenkoalexey (changes requested):

jack8558 (approved):

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,155 @@
+"""
+Activation checkpointing, scan compilation, and optimization barrier injection utilities.
```
yaoshiang, line 2: Please copy the format here: 

https://google.github.io/styleguide/pyguide.html#382-modules

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,155 @@
...
+
+This module defines utilities to apply memory-saving techniques during model training,
+specifically for use with PyTorch/XLA and SPMD sharding. It includes logic to:
+
+- Add activation rematerialization (checkpointing) to selected layers.
+- Compile repeated modules with `scan` for memory/compute efficiency.
+- Enable host tensor offloading during checkpointing.
+- Apply backward optimization barriers to mitigate recompute overhead.
+
+Functions:
```
yaoshiang, line 12: Don't need to redocument functions in the module docstring

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,155 @@
...
+- `add_activation_checkpointing_and_scan`: Configures checkpointing and scan transforms.
+- `add_optimization_barriers`: Inserts XLA backward barriers into specified layers.
+- `_get_classes_by_names`: Maps string layer names to actual model class objects.
+"""
+
+import logging
+from functools import partial
+
+import torch.nn as nn
```
yaoshiang, line 21: Please lint with ruff to ensure imports are sorted and organized.

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,155 @@
...
+    else partial(
+      offloading.remat_all_and_offload_these_inputs, names_to_offload=offload_tensors
+    )
+  )
+  return scan_layers.compile(model, layers_to_scan, partition_fn=partition_fn)
+
+
+def add_optimization_barriers(model: nn.Module, config: DictConfig) -> nn.Module:
+  """
+  Applies backward optimization barriers to specified layer types.
```
yaoshiang, line 104: Please follow the style from the style guide:

```python
"""One line description.

A blank line, then a longer description of the function.

Args:

Returns

You can skip some raises, see https://google.github.io/styleguide/pyguide.html#244-decision

File: torchprime/torch_xla_models/optimization/remat_and_scan.py [hunk 10]
```diff
@@ -0,0 +1,155 @@
...
+    torch.nn.Module: Modified model with optimization barriers.
+  """
+  remat_config = config.model.remat
+  classes = _get_classes_by_names(
+    model, remat_config.get("optimization_barrier_layers", [])
+  )
+  if not classes:
+    return model
+
+  logger.info(f"Adding backward optimization barriers to {classes}")
```
yaoshiang, line 123: Please update this to never use f strings for logs. For performance, logs should always be conditionally evaluated:

```python
logger.info('Adding... %s', classes)
```

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def _get_classes_by_names(
+  model: nn.Module, class_names: list[str]
+) -> tuple[type[nn.Module], ...]:
+  """
+  Helper to resolve string class names to actual model classes.
+
+  Args:
+    model (torch.nn.Module): Reference model to resolve context.
```
yaoshiang, line 140: do not need to repeat type here since the typehint covers it

yaoshiang (changes requested):
I left some spot comments but they are not comprehensive - please apply to PR and get another review from readability. If it's helpful, review the Google Python Style Guide.

torchprime/torch_xla_models/tests/test_trainer_cpu.py
```diff
@@ -0,0 +1,174 @@
...
+@patch(
+  "torchprime.torch_xla_models.sharding.initialization.get_mesh",
+  return_value=FakeMesh(),
+)
+@patch(
+  "torchprime.torch_xla_models.sharding.initialization.shard_torch_xla_model_from_config",
+  side_effect=lambda model, *args, **kwargs: model,
+)
+@patch("torchprime.torch_xla_models.trainer.basic.xm.xla_device", return_value="cpu")
+@patch("torchprime.torch_xla_models.trainer.basic.torch_xla.sync")
```
vlasenkoalexey, line 126: Looks like you are mocking out all PyTorch/XLA specific logic, is there any reason to do so?
It limits the usefulness of such test, and this test is really useful, thanks for putting it together.
FYI, majority of tests in torchrpime is executed on CPU, without mocking any logic.
Would it be possible to make it working without mocking?
If not, please add a comment describing why you think such setup is not possible.

vlasenkoalexey: I don't see why we shouldn't, this is a great test. Could you at least give it a try, and if doesn't work document problem

vlasenkoalexey (approved):
One ask is to see if it is possible to remove mocks from trainer test. If not possible for some reason, please add comments why.

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,152 @@
...
+- Apply backward optimization barriers to mitigate recompute overhead.
+"""
+
+import logging
+from functools import partial
+
+import torch.nn as nn
+import torch_xla.distributed.spmd as xs
+from omegaconf import DictConfig
+from torch_xla.distributed.fsdp import checkpoint_module
```
yaoshiang, line 18: Not your fault here but modules should never be named "module"

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,152 @@
+"""Activation checkpointing, scan compilation, and optimization barrier injection utilities.
```
yaoshiang, line 1: There's inconsistent usage of "activation rematerialization" vs "activation checkpointing" in this file. Given a goal was to eliminate potential confusion from the multiple uses of the term "checkpoint", I'd recommend removing all reference to "checkpoint" other than mentioning that "activation rematerialization" is aka "activation checkpointing".

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,152 @@
...
+
```
yaoshiang, line 2: this package name can be confusing - optimization is also the term for backprop.

torchprime/torch_xla_models/optimization/remat_and_scan.py
```diff
@@ -0,0 +1,152 @@
...
+This module defines utilities to apply memory-saving techniques during model training,
+specifically for use with PyTorch/XLA and SPMD sharding. It includes logic to:
```
yaoshiang, line 4: this module name could be improved - remat_and_scan both uses an abbreviation and has an "and". 

Google Python Style Guide: "Avoid abbreviation". 

Is there a better package and module name we can use here to describe what these do. Is this purely about memory efficiency during training? Scan is about XLA efficiency right? If this is about rewriting the nn.module for training efficiency.. maybe something like 

torchprime.torch_xla_models.efficiency.rewrite.py? 

I'm not in love with that phrase but it kind of gets the message across a bit better. Open to a discussion on this.

torchprime/torch_xla_models/sharding/initialization.py
```diff
@@ -0,0 +1,69 @@
...
+`nn.Linear` are patched for SPMD compatibility and leverages user-provided configuration
+(OmegaConf) to control sharding behavior.
+"""
+
+import logging
+
+import torch.nn as nn
+import torch_xla.distributed.spmd as xs
+from omegaconf import DictConfig, OmegaConf
+from torch_xla.distributed.spmd.xla_sharding import apply_xla_patch_to_nn_linear
```
yaoshiang, line 14: is ruff saying these imports are sorted?

torchprime/torch_xla_models/sharding/initialization.py
```diff
@@ -0,0 +1,69 @@
+"""Sharding initialization module for TPU-based training using PyTorch/XLA SPMD APIs.
```
yaoshiang, line 1: this this just be a module called sharding.py, rather than a singleton package with one module, which itself has one key function?

yaoshiang (changes requested):
I think the code is looking much better! Nice. I'm going to nit on the naming of this package given this is meant to be a clean demonstration code. let's think about the right package and module names here. Open to a group discussion.

torchprime/torch_xla_models/sharding/initialization.py
jialei777: yeah it is the result of ruff format. I think it sort import xxx and from xxx separately

torchprime/torch_xla_models/sharding/initialization.py
jialei777: there is also another sharding.py file outside torch_xla_model folder. I think we need to clean up things later

torchprime/torch_xla_models/optimization/remat_and_scan.py
jialei777: okay, change the name into remat_utils.py

torchprime/torch_xla_models/optimization/remat_and_scan.py
jialei777: consolidate files into `model_rewriting` folder

yaoshiang (commented):
marking this as a "comment" to unblock merge. I continue to think that the naming and documentation could use improvement for clarity for readers.

yaoshiang (commented):
leaving a "comment" to unblock merge. I continue to think that the naming of packages and modules, and inconsistencies in the docstrings, will make it harder for readers to get value out of the high quality code contained here.

yaoshiang (approved):
approving to unblock merge. I continue to think there's more we can do to make the package, module, and documentation of this package and module to be easier for users to consume.

PR #266, opened 2025-05-30
Support tensorboard metrics in the torchprime trainer

tengyifei:
This PR adds tensorboard support to the `torch_xla_models` trainer and document it.

The primary motivation is to store loss curves for subsequent analysis. But this should be useful for modellers and researchers as well.


torchprime/torch_xla_models/trainer/base_trainer.py
```diff
@@ -163,6 +181,12 @@ def _get_train_dataloader(self) -> pl.MpDeviceLoader:
     )
     return loader
 
+  def _log_to_tensorboard(self, step: int, loss: float, learning_rate: float):
+    """Log metrics to TensorBoard."""
+    self.summary_writer.add_scalar("train/loss", loss, step)
```
jialei777, line 186: Add epoch also? something like, `curr_step/len(train_dataset)`

jialei777 (approved):

torchprime/torch_xla_models/configs/default.yaml
```diff
@@ -7,6 +7,8 @@ defaults:
   - _self_ # refers to this config file
   - model: llama-3-8b # refers to model/llama-3-8b.yaml
 
+# The name of the training run as it shows up on tensorboard.
+run_name: null
```
vlasenkoalexey, line 11: should it be run_name: default ?

docs/tensorboard.md
```diff
@@ -0,0 +1,29 @@
...
+
+```sh
+torchprime/torch_xla_models/train.py run_name=my-exp logging_steps=1
+```
+
+The metrics contains `loss` and `learning_rate` etc. and you can visualize their
+progression by starting a tensorboard web server:
+
+```sh
+tensorboard --logdir outputs/runs
```
vlasenkoalexey, line 22: nit: add --port=6006

docs/tensorboard.md
```diff
@@ -0,0 +1,29 @@
+# How to view metrics on tensorboard
```
vlasenkoalexey, line 1: I assume that same should work on cluster using tp run.
Could you call it out here as well

vlasenkoalexey (approved):

PR #275, opened 2025-06-02
Distributed Checkpointing

hlnchen:
3 config added to control checkpoints:

- `checkpoint_dir`: directory of the checkpoint, use gs://bucket/path/to/checkpoint to save to gcs bucket
- `resume_from_checkpoint`:
  - if `null`, will not load checkpoint but load from huggingface pretrained weights
  - if a positive integer step, checkpoint manager will try to find and load weights from checkpoint under `checkpoint_dir/resume_from_checkpoint/`
  - if `latest` or the step not found by the manager, then last checkpoint will be loaded
- save_steps: save frequency
- checkpoint state dict
  - model
  - optimizer
  - lr_scheduler
  - step
 
after checkpoint loading will skip first `step` iterations by looping the dataloader.

What's not included:
- async checkpointing
- saving in Huggingface format

### DIFF
--- a/torchprime/torch_xla_models/configs/default.yaml
+++ b/torchprime/torch_xla_models/configs/default.yaml
@@ -26,7 +26,7 @@ output_dir: outputs
 
 # Checkpoint directory
 checkpoint_dir: checkpoints/
-checkpoint_step: null
+resume_from_checkpoint: null
 save_steps: 15
 
 optimizer:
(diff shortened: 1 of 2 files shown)


hlnchen:
@yaoshiang actually there is something we are not sure: when using `torch.distributed.checkpoint` or `CheckpointManager`, do we need to save/load the model on devices with the same type of device (so that sharding are the same)? Also, it doesn't look like to be compatible with `https://github.com/pytorch/xla/blob/master/torch_xla/distributed/fsdp/state_dict_utils.py#L109`. Any suggestion to consolidate the distributed checkpoint into a single one, like HF model checkpoint?

PR #279, opened 2025-06-03
Refactor model class and add from_pretrained and export

jialei777:
Add a `BaseCausalLM` model class.
Add unit test for 
- loading models <10B (inference) parameters: `test_model_utils.py`
- loading a small llama 1b from hf to avoid CI OOM: `test_load_model_from_hf.py`
- saving and reloading a small llama model with same forward logics: `test_model_loading_saving.py`


torchprime/torch_xla_models/model/base_causal_lm.py
```diff
@@ -0,0 +1,96 @@
...
+
+import torch
+import torch.nn as nn
+from huggingface_hub import snapshot_download
+from omegaconf import OmegaConf
+from safetensors import safe_open
+from safetensors.torch import save_file
+
+
+def load_sharded_safetensors_to_state_dict(model_dir: str) -> dict:
```
vlasenkoalexey, line 19: as discussed, load can happen on all replicas, but save should happen only on a first one.

please use proper docstrings with args and returns

torchprime/torch_xla_models/model/base_causal_lm.py
```diff
@@ -0,0 +1,96 @@
...
+    if isinstance(module, nn.Linear):
+      module.weight.data.normal_(mean=0.0, std=std)
+      if module.bias is not None:
+        module.bias.data.zero_()
+    elif isinstance(module, nn.Embedding):
+      module.weight.data.normal_(mean=0.0, std=std)
+      if module.padding_idx is not None:
+        module.weight.data[module.padding_idx].zero_()
+
+  def forward(
```
vlasenkoalexey, line 64: proper docsting here and below please

torchprime/torch_xla_models/model/model_utils.py
```diff
@@ -0,0 +1,65 @@
...
+  if not hasattr(module, model_class_name):
+    print(f"Error: Class '{model_class_name}' not found in module '{module.__name__}'")
+    sys.exit(1)
+
+  model_class = getattr(module, model_class_name)
+  return model_class(model_config)
+
+
+@contextmanager
+def set_default_dtype(dtype):
```
vlasenkoalexey, line 36: docstring please

torchprime/torch_xla_models/tests/test_model_utils.py
```diff
@@ -0,0 +1,53 @@
...
+import yaml
+from omegaconf import OmegaConf
+
+from torchprime.torch_xla_models.model.base_causal_lm import BaseCausalLM
+from torchprime.torch_xla_models.model.model_utils import (
+  extract_model_size_from_model_name,
+  initialize_model_class,
+)
+
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../model")))
```
vlasenkoalexey, line 26: why is this necessary, and can we do without it?
If we have to do it, this should have a comment explaing why this is needed

torchprime/torch_xla_models/tests/test_load_model_from_hf.py
```diff
@@ -0,0 +1,54 @@
...
+import sys
+
+import pytest
+import yaml
+from omegaconf import OmegaConf
+
+from torchprime.torch_xla_models.model.base_causal_lm import BaseCausalLM
+from torchprime.torch_xla_models.model.model_utils import initialize_model_class
+
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../model")))
```
vlasenkoalexey, line 22: same q

.github/workflows/cpu_test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
   pytest:
     needs: setup
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-22.04-16core
```
vlasenkoalexey, line 31: why is this needed?

vlasenkoalexey (changes requested):

torchprime/torch_xla_models/model/base_causal_lm.py
```diff
@@ -0,0 +1,96 @@
...
+  weight_map = index["weight_map"]
+  for filename in set(weight_map.values()):
+    path = os.path.join(model_dir, filename)
+    with safe_open(path, framework="pt", device="cpu") as f:
+      for key in f.keys():  # noqa: SIM118
+        state_dict[key] = f.get_tensor(key)
+  return state_dict
+
+
+def save_sharded_safetensors_by_layer(state_dict: dict, save_dir: str):
```
vlasenkoalexey, line 34: if we want to consume produced checkpoint in HF, is there any other metadata or config that we should export or copy from original checkpoint?

How is it usually done?

.github/workflows/cpu_test.yml
jialei777: I am actually loading a llama-8b into memory in `torchprime/torch_xla_models/tests/test_load_model_from_hf.py` and seems it failed with OOM with ubuntu-22.04. It works in my local machine, so want to check if increase vm size can help.

torchprime/torch_xla_models/model/base_causal_lm.py
jialei777: Model checkpoint is literately just a dict from tensor name to tensor, so I think there is no additional config apart from weight_map.

We also write our model config in the export method just for recording configs.

torchprime/torch_xla_models/tests/test_model_utils.py
jialei777: I think at this point we cannot remove it. But let me fix it.

jialei777: fixed

Issue #90, opened 2025-02-06
[torch_xla] MVP correctness and convergence check for Llama 3.0 8B


tengyifei:
- Grab a log of a Hugging Face Llama 3.0 8B run. Example: Llama 3.0 8B with gbs 256 on 1 Trillium pod.
  + Initialize the model with some fixed seed
  + Train the model for 50 steps
  + Record the loss at every step
  + At the end of the training, record the output of the model on some example tokens
  + Save this as a reference file

- Create a test of training Llama 3.0 8B with torchprime with the same configs
- Use the same seed to initialize the model
- Compare the loss against the HF reference at every iteration
- Compare the final output with that from HF
- We should run this test at least in post submit CI

zpcore:
I have the loss for 500 steps with the following:

Configuration:
```
llama 3 8B, model fsdp sharding over 256
8K sequence length
training dataset: wikitext-103-raw-v1
lr: 5.e-5
max_step: 500
minibatch on
Sampler: torch.utils.data.DistributedSampler
```
HF code snapshot [link](https://github.com/pytorch-tpu/transformers/tree/piz/correctness_verify_loss) for generating data.

<img width="973" alt="Image" src="https://github.com/user-attachments/assets/7da5d75f-9b0d-4766-b66d-55dae1e6b337" />

The concern is that TP doesn't reach the same loss as what HF can provide. I will check more detail on the optimizer and data representations to see if there are any difference.

tengyifei:
That's a good catch

tengyifei:
Maybe we need to drop the last batch from the data loader? (HF config had a `dataloader_drop_last: true` feature, which says to discard the last potentially incomplete batch).

If the batch size is too small for the last batch, maybe that could cause the loss spikes in the TP run.

zpcore:
> Maybe we need to drop the last batch from the data loader? (HF config had a `dataloader_drop_last: true` feature, which says to discard the last potentially incomplete batch).
> 
> If the batch size is too small for the last batch, maybe that could cause the loss spikes in the TP run.

Right, I think this can be different, but when I check the epoch number and step number, HF run and TP run matches, which means they reset the dataloader iterator at the same time and start a new epoch. Below are the last few loss data entries from HF and TP:

HF:
```
[INFO\|trainer.py:2295] 2025-02-12 21:41:02,193 >> Running Epoch: 37, Step 491, tr_loss_step: 4.584621429443359
[...]

PR #286, opened 2025-06-06
Implement gradient clipping

tengyifei:
Fixes https://github.com/AI-Hypercomputer/torchprime/issues/90

The Hugging Face trainer clips gradient by norm by default. It turns out this makes a big difference in training stability and simply adding gradient clipping will bring torchprime into parity with Hugging Face.

Experiment: http://tb/share/yuBUPRF6KqZKPhwJkXNvc

screenshot-2025-06-05-21-54-18

Tested:

```
tp run --name tp-linear-clip-norm-2 torchprime/torch_xla_models/train.py model=llama-3-8b ici_mesh.fsdp=256 profile_step=3 profile_duration=30000 task.max_steps=1000 logging_steps=1 task.global_batch_size=256 dataset.hf_dataset_config_name=wikitext-103-raw-v1 run_name=tp-linear-clip-norm-2

tp run --use-hf torchprime/hf_models/train.py train_script.args.per_device_train_batch_size=256 +train_script.args.log_loss=true train_script.args.logging_strategy=steps +train_script.args.logging_steps=1 +train_script.args.logging_first_step=true +train_script.args.report_to=tensorboard train_script.args.max_steps=1000
```


yaoshiang (changes requested):
grad clipping by norm is not the only way to clip grads. Clipping by absolute value is valid. Can you update this PR to enable both? Also, the grad clipping should be configuring outside the optimizer. The grad clipping is upstream of the optimizer - the optimizer has no knowledge of grad clipping - can you move the grad clipping config to a top-level section?

jialei777 (approved):

yaoshiang (approved):
looking good! I realize you are pushing hard to finish this off, if you have cycles, please add a unit test of some kind. Numerical would be ideal but might take a bit of time. You could do like a single layer, with fixed numbers, have the loss function be MAE, so now you know the exact grads, try your clip, and see if you get what you expect. If you feed those instructions to an LLM it'll probably do a decent job. But consider it optional since you have numerical proof against the HF curves that the norm version at least is working./ Thanks!

PR #308, opened 2025-06-13
Support skipping tracing of selected pure modules

tengyifei:
This PR adds a `model.pure_modules` option to the trainer, which indicates which modules are run with `@assume_pure` [1].

The primary benefit is making available rich profiles during the backward pass, because `jax.vjp` preserves framework scopes in the backward pass.

This PR updates PyTorch/XLA pin to Jun 17 because it relies on https://github.com/pytorch/xla/pull/9360.

## Regular profile

```
python3 torchprime/torch_xla_models/train.py \
    model/sharding=llama-fsdp-tp ici_mesh.fsdp=8 \
    task.global_batch_size=8 model.attention_kernel=splash_attention \
    logging_steps=1 task.max_steps=15
```

screenshot-2025-06-17-14-15-31

## Profile with `assume_pure`

```
python3 torchprime/torch_xla_models/train.py \
    model/sharding=llama-fsdp-tp ici_mesh.fsdp=8 \
    task.global_batch_size=8 model.attention_kernel=splash_attention \
    logging_steps=1 task.max_steps=15\
    model.pure_modules=[LlamaMLP,EinsumLinear]
```

screenshot-2025-06-17-14-15-20

[1]: https://docs.pytorch.org/xla/master/perf/assume_pure.html

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -100,6 +100,7 @@ jobs:
             dataset=wikitext \
             task=train \
             task.global_batch_size=8 \
+            task.lr_scheduler.type=constant \
             task.max_steps=15 \
             ici_mesh.fsdp=4 \
             profile_start_step=3 \
--- a/e2e_testing/step_time_bounds.yaml
+++ b/e2e_testing/step_time_bounds.yaml
@@ -1,54 +1,63 @@
 benchmarks:
   llama-3-8b:
     name: Llama 3.0 8B
-    step_time_lower_bound: 0.894678
-    step_time_upper_bound: 4.54563437
-    confidence_interval: 1.82548
-    average: 2.7202
-    sample_size: 114
+    step_time_lower_bound: 2.68109009
+    step_time_upper_bound: 2.789223
+    confidence_interval: 0.05407
+    average: 2.7352
+    sample_size: 427
+  llama-3-8b-pure-mlp:
+    name: Llama 3.0 8B (@assume_pure)
+    # Bounds are copied from `llama-3-8b`. They will be overwritten the next time
+    # somebody runs `e2e_testing/update_step_time.py`.
+    step_time_lower_bound: 2.68109009
+    step_time_upper_bound: 2.789223
+    confidence_interval: 0.05407
+    average: 2.7352
+    sample_size: 1
   llama-3_1-8b-sa:
     name: Llama 3.1 8B (Splash Attention)
-    step_time_lower_bound: 2.35428493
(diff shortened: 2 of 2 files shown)


torchprime/torch_xla_models/model_rewriting/assume_pure.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def _pure_forward(module, params, buffers, args, kwargs):
+  return torch.func.functional_call(module, (params, buffers), args, kwargs)
+
+
+def replace_nn_linear_with_einsum(module: torch.nn.Module, config: DictConfig):
+  """Recursively replace `nn.Linear` layers with `EinsumLinear` in the module.
+
+  Without this patch, an `nn.Linear` module in PyTorch/XLA will lower to reshapes
```
yaoshiang, line 64: I recall that xla's einsum worked fine when used without custom-op - can you point me to where the einsum becomes a custom op and remind me why?

do we need to use this `torch_xla.distributed.spmd.xla_sharding.apply_xla_patch_to_nn_linear` - could we use assume_pure on the linear model directly in this code? That question may be missing some basic knowledge of assume-pure, my apologies if so.

torchprime/sharding/shard_model.py
```diff
@@ -197,7 +197,7 @@ def shard_param(tensor, spec: tuple[str, ...]):
     # and models are usually constructed eagerly in torchax.
     return torch_view(
       jax.make_array_from_callback(
-        tensor.shape, sharding, lambda slice_index: tensor[slice_index]
+        tensor.shape, sharding, lambda slice_index: jax_view(tensor[slice_index])
```
vlasenkoalexey, line 200: for my understanding, why is this needed, and is there an impact on performance?

vlasenkoalexey (approved):
thanks for adding this functionality, this is great

torchprime/sharding/shard_model.py
tengyifei: Ah, this was just a bug. It probably used to work accidentally.

This code shards the model weights for torchax models. It does so by indexing into an unsharded torch tensor and giving views into that tensor to the `jax.make_array_from_callback` function.

The `jax.make_array_from_callback` function only works with JAX arrays. So the right thing to do is to use `jax_view` to wrap the torch tensor into a jax array before giving it to jax. It looks like the newer JAX version that comes from the updated PyTorch/XLA pin has enforced stricter checks that surfaced this bug.

In any case, none of this applies to `torch_xla` models and I had to fix the `torchax` specific tests as part of updating to PyTorch/XLA Jun 17.

PR #322, opened 2025-06-24
Write HF config file/tokenizer to output folder for vllm deployment 

jialei777:
- copy Hugging Face `config.json` and `generation_config.json` when loading checkpoints
- export tokenizer files with configs

Then sft-ed ckpts can be directly loaded with vllm with [this script](https://github.com/AI-Hypercomputer/torchprime/pull/314/files#diff-aa9e4c1175fe80fc934171688814b16ddb6a676ec1a302af389dfefae175a533)

- remove the distcp ckpts if we decide to convert back to safetensors for deployment

This is to save some disk space.

### DIFF
--- a/torchprime/torch_xla_models/model/base_causal_lm.py
+++ b/torchprime/torch_xla_models/model/base_causal_lm.py
@@ -100,8 +100,10 @@ class BaseCausalLM(nn.Module):
     with open(os.path.join(save_directory, "config.json"), "w") as f:
       json.dump(OmegaConf.to_container(self.config, resolve=True), f, indent=2)
 
-  def from_pretrained(self, model_path_or_repo: str):
-    """Load model weights from local directory or Hugging Face Hub repo.
+  def from_pretrained(
+    self, model_path_or_repo: str, *, save_dir: str | os.PathLike | None = None
+  ) -> None:
+    """Load model weights and optionally save configs/tokenizer.
 
     This method loads the model's state dictionary from the specified path or repository.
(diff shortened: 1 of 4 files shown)


liurupeng:
LGTM

liurupeng:
waiting for @vlasenkoalexey to take one more look as well

torchprime/torch_xla_models/model/base_causal_lm.py
```diff
@@ -100,8 +100,10 @@ def export(self, save_directory: str):
     with open(os.path.join(save_directory, "config.json"), "w") as f:
       json.dump(OmegaConf.to_container(self.config, resolve=True), f, indent=2)
 
-  def from_pretrained(self, model_path_or_repo: str):
-    """Load model weights from local directory or Hugging Face Hub repo.
+  def from_pretrained(
```
vlasenkoalexey, line 103: it doesn't look like a right API design to have from_pretrained to take save_directory and doing some saving.

There are 2 ways:
- from_pretraining to store some metadata internally about config and tokenizer and saves it along with other files in save_checkpoint (guarded by additional argument)
- Add new save_huggingface_configs method that downloads and saves configs

vlasenkoalexey (changes requested):

torchprime/torch_xla_models/model/base_causal_lm.py
jialei777: let me move saving config to a separate save_huggingface_configs method

vlasenkoalexey (approved):
thank you, looks much more logical now

PR #331, opened 2025-07-01
Add loss check for llama-3-8b-sft workflow

jialei777:
## Summary
- introduce `check_loss.py` for validating final loss from logs
- extend reusable_e2e_check workflow to support optional loss validation
- propagate new fields through e2e_test.yml
- document target loss and tolerance in step_time_bounds
- run SFT benchmark for 50 steps

## Testing
- `pytest -q` *(fails: ModuleNotFoundError: No module named 'torchprime')*

------
https://chatgpt.com/codex/tasks/task_e_6864018ce7f0832c8ff76d973a1f2692

### DIFF
--- a/.github/workflows/reusable_e2e_check.yml
+++ b/.github/workflows/reusable_e2e_check.yml
@@ -19,6 +19,14 @@ on:
         description: "Upper bound for step time (in seconds)"
         required: true
         type: number
+      target_loss:
+        description: "Expected final loss value"
+        required: false
+        type: string
+      loss_tolerance:
+        description: "Allowed deviation from the expected loss"
+        required: false
+        type: string
     secrets:
       GCP_SA_KEY:
         required: true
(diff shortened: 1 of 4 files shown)


.github/workflows/reusable_e2e_check.yml
```diff
@@ -19,6 +19,14 @@ on:
         description: "Upper bound for step time (in seconds)"
         required: true
         type: number
+      target_loss:
+        description: "Expected final loss value"
+        required: false
+        type: string
```
vlasenkoalexey, line 25: why is it a string?

e2e_testing/check_loss.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    return 1
+
+  print("Loss check passed.")
+  return 0
+
+
+if __name__ == "__main__":
+  if len(sys.argv) != 4:
+    print("Usage: check_loss.py <log_file> <target_loss> <tolerance>")
```
vlasenkoalexey, line 49: nit could you add this to the module doc.
No need to change anything here, just FYI that usually we use absl flags to pass data

vlasenkoalexey (approved):
Thank you!

.github/workflows/reusable_e2e_check.yml
jialei777: updated with number

PR #332, opened 2025-07-01
Jackoh/resnet18 analysis

jack8558:
Draft for integrating Resnet-18 and adding analysis.

Currently has original analysis of @yaoshiang and added some README for analysis.

PR #335, opened 2025-07-09
Add ResNet-18 Vision Model and Training Example

jack8558:
This PR introduces an initial implementation of the ResNet18 model from thus [analysis](http://shortn/_Bbyr9ZwoaB).

A few points on its current state:

- We're adding the model and registering it in model.registry, following the proposal in [AI-Hypercomputer/torchprime#216](https://github.com/AI-Hypercomputer/torchprime/pull/216).
- For now, train.py, data.py, and metrics_log.py are placed in torch_xla_vision. This is a temporary home as we're still exploring how best to integrate this example which are beyond LLM.
- Using the CelebA dataset instead of FaceID-550+vggface as the equivalent dataset in huggingface was not found. 
- Doing binary image classification as CelbA is not best suit for face classification for insufficient data for each classes.

I recognize this PR introduces new components that may not yet perfectly align with our broader architecture. My aim is to provide a working foundation; let's integrate this now and refine it collaboratively as our vision for torchprime evolves with more diverse models. Open for change and feedbacks!


pyproject.toml
```diff
@@ -30,16 +30,19 @@ dependencies = [
...
 [project.optional-dependencies]
 dev = [
+    "fire==0.7.0",
     "ruff~=0.11.4",
     "pytest~=8.3.4",
     "pytest-forked~=1.6.0",
     "toml~=0.10.2",
     "dataclasses-json~=0.6.7",
     "watchdog~=6.0.0",
+    "gdown~=5.2.0", # For torchvision CelebA dataset download
```
yaoshiang, line 40: possible to remove this now that we are using HF?

We have too many deps in general so we actually want to go the other direction here... the more deps ,the harder it is to repro and understand.

pyproject.toml
```diff
@@ -30,16 +30,19 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
+    "fire==0.7.0",
```
yaoshiang, line 33: we are already using omegaconf, click, and hydra... can we reuse one of those instead of adding another CLI flag manager?  We might be using absl as well, not sure.

torchprime/models/resnet18/__init__.py
```diff
@@ -0,0 +1,24 @@
+"""ResNet-18 model package."""
+
+import torch
+
+from .model import Resnet18ForClassification
+
+
+def get_model(
```
yaoshiang, line 8: Please check with @vlasenkoalexey on including factories, and, logic in `__init__.py` files. I defer to him on best practice / google style.

yaoshiang (approved):
approved pending responses to comments

pyproject.toml
jack8558: Sound good, using hydra like torch_xla_models/train.py now.

torchprime/models/resnet18/__init__.py
jack8558: The factories and __init__.py files are there because I was following the approach in [PR #216](https://github.com/AI-Hypercomputer/torchprime/pull/216). The idea is to register all models in the /models directory within a registry, which makes them easier to reuse across different scripts. For example, the Transact model added in that PR is used in run_model_forward.py for e2e testing.

I went with this proposal for a couple of reasons:
- The Transact model was added for a similar purpose as Resnet18—it was a customer-related piece of work we wanted to reference, and it's different from our other LLMs.
- I thought it was a good idea to have all models in one central registry; it should make it simpler to reuse them as we add various scripts, like different types of training scripts.

I'm not super rigid about sticking to this paradigm, though. If we decide not to go this route, I'm happy to follow whatever you think is best.

torchprime/torch_xla_vision/train.py
```diff
@@ -0,0 +1,294 @@
...
+    if config.use_torch_compile:
+      logger.info("Using torch.compile on CUDA.")
+      compile_fn = torch.compile
+    else:
+
+      def compile_fn(fn):
+        return fn
+
+  else:
+    os.environ["XLA_COMPILE_CACHE_PATH"] = "/tmp/xla_cache"
```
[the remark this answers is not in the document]
jack8558: Modified based on the doc provided in previous comment: google3/third_party/torch_tpu/examples/mingpt/mingpt_example.py

IIUC, the purpose is to cache compiled graph for speedup but lmk if we would like to not include this part.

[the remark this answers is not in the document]
jack8558: I think this is actually introducing some warnings at least in my case in TPU VM.

`2025-07-12 01:17:32.208079: W torch_xla/csrc/runtime/pjrt_computation_client.cpp:680] Failed to deserialize executable: UNIMPLEMENTED: Deserializing serialized executable not supported.`

Should we remove this part if it's not needed? @yaoshiang

yaoshiang:
> > This PR doesn't yet include a viz.ipynb right? Just the training? How is the training on GPU vs TPU with the same hyperparameters?
> 
> Yes, I am planning to include viz.ipynb and the Readme in separate PR at: #332
> 
> The analysis has to be updated since there were some changes such as dataset (to CelebA from VggFace). I will update the viz.ipynb and README once this PR is merged so I can use the actual training example which is submitted. I am also currently running the training on GPU and will update here the result shortly.

sounds great, was just checking my understanding - thank you.

jack8558:
Using the current default hyperparameters, the result is below.

GPU A100: 
`[2025-07-15 19:33:00,742][__main__][INFO] - {'accuracy_top1': '0.3234', 'test_loss': '4.9365', 'train_loss': '1.6234'}`

TPU v4-8:
`[2025-07-15 19:03:59,157][__main__][INFO] - {'accuracy_top1': '0.3709', 'test_loss': '4.6414', 'train_loss': '1.1325'}`

vlasenkoalexey (approved):

torchprime/torch_xla_vision/train.py
```diff
@@ -0,0 +1,296 @@
...
+      def compile_fn(fn):
+        return fn
+
+  else:
+    import torch_xla
+
+    # Initialize the XLA compilation cache. This saves compiled graphs to disk,
+    # which significantly speeds up the startup time for subsequent runs.
+    # https://docs.pytorch.org/xla/release/r2.7/learn/pytorch-on-xla-devices.html#compilation-caching
+    torch_xla.runtime.initialize_cache("/tmp/xla_cache", readonly=False)
```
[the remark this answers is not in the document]
jack8558: Removed this part due to multiple unexpected warnings 
https://github.com/AI-Hypercomputer/torchprime/pull/335#discussion_r2202223210

PR #336, opened 2025-07-10
upgrade transformers version to 4.52.1

jialei777:


vlasenkoalexey:
Mainly for my understand, what was the reason to bump up version?
LGTM assuming it doesn't break our tests.

jialei777:
> Mainly for my understand, what was the reason to bump up version? LGTM assuming it doesn't break our tests.

It is suggested by the dependency bot https://github.com/AI-Hypercomputer/torchprime/pull/334

PR #339, opened 2025-07-12
adding Llama4 Scout config, and making it to work without OOM

vlasenkoalexey:
Using a naive slow implementation for MOEs, the right way to fix this would be by implementing Expert Parallelism.

Confirmed that can train Llama-4-Scout-17B-16E text-only model now:
```sh
tp run --name av-llama-4-scout-17b-16e \
    model=llama-4-text-only-scout-17b-16e \
    task.global_batch_size=256 \
    ici_mesh.fsdp=64 \
    ici_mesh.tensor=4 \
    dataset.block_size=512
```

### DIFF
--- a/torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
+++ b/torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
@@ -1,0 +1,149 @@
+# Modelled by
+# https://huggingface.co/meta-llama/Llama-4-Scout-17B-16E/blob/main/config.json
+defaults:
+  - _self_  # refers to this config file
+  - sharding: llama4-fsdp-tp  # refers to sharding/llama4-fsdp-tp.yaml
+  - remat: llama4  # refers to remat/llama4.yaml
+
+model_class: llama4.Llama4TextForCausalLM  # Used to import the model from this class
+tokenizer_name: meta-llama/Llama-4-Scout-17B-16E
+torch_dtype: bfloat16
+use_bfloat16: true
+temperature: 1.0
+bos_token_id: 200000
+pad_token_id: 200018
+eos_token_id:
+- 200001
+- 200007
+- 200008
+attention_bias: false
+model_type: llama4_text
+attn_temperature_tuning: 4  # not sure if needed
+attn_scale: 0.1
+floor_scale: 8192
+vocab_size: 202048
+max_position_embeddings: 262144
+hidden_size: 5120
+intermediate_size: 8192
+intermediate_size_mlp: 16384
+num_hidden_layers: 48
+num_attention_heads: 40
+rope_scaling:
+  factor: 16.0
+  high_freq_factor: 1.0
+  low_freq_factor: 1.0
+cache_implementation: hybrid # not sure if needed
+num_key_value_heads: 8
+hidden_act: silu
+initializer_range: 0.02
+rms_norm_eps: 1.0e-05
+use_cache: true      # not sure if needed
+rope_theta: 500000.0
+attention_dropout: 0.0
+head_dim: 128
+use_qk_norm: true
+num_experts_per_tok: 1
+num_local_experts: 16
+output_router_logits: false
+router_aux_loss_coef: 0.001
+router_jitter_noise: 0.0
+no_rope_layers:
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+moe_layers:
+- 0
+- 1
+- 2
+- 3
+- 4
+- 5
+- 6
+- 7
+- 8
+- 9
+- 10
+- 11
+- 12
+- 13
+- 14
+- 15
+- 16
+- 17
+- 18
+- 19
+- 20
+- 21
+- 22
+- 23
+- 24
+- 25
+- 26
+- 27
+- 28
+- 29
+- 30
+- 31
+- 32
+- 33
+- 34
+- 35
+- 36
+- 37
+- 38
+- 39
+- 40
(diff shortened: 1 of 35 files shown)


torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
```diff
@@ -0,0 +1,149 @@
...
+rope_theta: 500000.0
+attention_dropout: 0.0
+head_dim: 128
+use_qk_norm: true
+num_experts_per_tok: 1
+num_local_experts: 16
+output_router_logits: false
+router_aux_loss_coef: 0.001
+router_jitter_noise: 0.0
+no_rope_layers:
```
jialei777, line 50: why the official LLAMA4 config don't have this?

torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
```diff
@@ -0,0 +1,149 @@
...
+- 0
+- 1
+- 1
+- 1
+- 0
+- 1
+- 1
+- 1
+- 0
+moe_layers:
```
jialei777, line 99: same why https://huggingface.co/meta-llama/Llama-4-Scout-17B-16E/blob/main/config.json don't have this?

torchprime/torch_xla_models/configs/model/sharding/llama4-fsdp-tp.yaml
```diff
@@ -0,0 +1,39 @@
+# 2D (FSDP + TP) sharding configuration for Llama4 models.
+
+# Weights
+
+# vocab_size, hidden_size
+model.embed_tokens.weight: [fsdp, tensor]
+# hidden_size, num_attention_heads * head_dim
+model.layers.*.self_attn.q_proj.weight: [tensor, fsdp]
```
jialei777, line 8: why the order is different compared to the fsdp-only config below?

torchprime/torch_xla_models/configs/model/sharding/llama4-fsdp.yaml
```diff
@@ -1,17 +1,34 @@
...
 model.embed_tokens.weight: [fsdp, null]
+# hidden_size, num_attention_heads * head_dim
 model.layers.*.self_attn.q_proj.weight: [fsdp, null]
-model.layers.*.self_attn.k_proj.weight: [null, fsdp]
-model.layers.*.self_attn.v_proj.weight: [null, fsdp]
-model.layers.*.self_attn.o_proj.weight: [fsdp, null]
-model.layers.*.feed_forward.experts.gate_up_proj: [fsdp, null, null]
-model.layers.*.feed_forward.experts.down_proj:  [fsdp, null, null]
+# hidden_size, num_key_value_heads * head_dim
+model.layers.*.self_attn.k_proj.weight: [fsdp, null]
```
jialei777, line 7: any particular reason for changing this?

torchprime/torch_xla_models/model/llama4/model.py
```diff
@@ -136,44 +136,105 @@ def __init__(self, config):
...
-    # this scales a lot better if you do EP!
+    # Now that we finished expert computation -> we scatter add because we gathered previously.
+    # W.e have to do this because we used all experts on all tokens.
     out.scatter_add_(dim=0, index=router_indices, src=routed_out.view(-1, hidden_dim))
     return out, router_scores
 
+  @xp.trace_me("Llama4TextMoe")
+  def forward(self, hidden_states):
+    """
+    This version processes experts sequentially to save memory. It is slower
```
jialei777, line 181: is the real solution to enable expert parallelism? Also if you keep increasing tensor parallelism dimension, it still OOM?

jialei777 (changes requested):
Can you also add an e2e example?

torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
vlasenkoalexey: It does have it implicitly, see https://github.com/huggingface/transformers/blob/967045082faaaaf3d653bfe665080fd746b2bb60/src/transformers/models/llama4/configuration_llama4.py#L300

This config is generated by running following snippet and manually removing stuff that we don't need:
config = AutoConfig.from_pretrained(
    "meta-llama/Llama-4-Scout-17B-16E"
)

torchprime_config = OmegaConf.create(config.to_dict())
torchprime_config

torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
vlasenkoalexey: See above comment

torchprime/torch_xla_models/configs/model/sharding/llama4-fsdp.yaml
vlasenkoalexey: good catch, this is a bug, mostly following this: https://github.com/AI-Hypercomputer/torchprime/blob/main/torchprime/torch_xla_models/configs/model/sharding/llama-fsdp.yaml

vlasenkoalexey:
Don't think model is ready to have an example yet, I added a sample training command to the PR description.

torchprime/torch_xla_models/model/llama4/model.py
jialei777: just curious, if you make tensor parallelism to 16. can it work?

torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
jialei777: Okay, then the question is why not we use the same `no_rope_layer_interval` parameter and make it implicitly? The current yaml file is difficult to read, and easy to miss the no rope layers. Also what about bigger models, say with 120 layers, then the yaml file will be supper long.

jialei777 (approved):

torchprime/torch_xla_models/configs/model/llama-4-text-only-scout-17b-16e.yaml
vlasenkoalexey: That's a good point, will make the change in a follow up PR.
One thing I'd like to explore is ability to reuse HF configs directly.

torchprime/torch_xla_models/model/llama4/model.py
vlasenkoalexey: I think I tried that, and it still didn't work for some reason. Didn't try to debug why.

PR #347, opened 2025-07-23
adding expert parallelism to Llama4

vlasenkoalexey:
Was able to run it as:
tp run --name alekseyv-llama-4-18-ep \
    model=llama-4-text-only-scout-17b-16e \
    task.global_batch_size=256 \
    ici_mesh.fsdp=16 \
    ici_mesh.data=1 \
    ici_mesh.expert=16 \
    dataset.block_size=512 \
    model/sharding=llama4-fsdp-tp-ep

Performance is still not great


jialei777 (approved):
Thank you for implementing this! It looks good to me, so approved the change. I have 3 comments tho:
1. Add e2e test maybe for a smaller version, so we can start collecting metrics to get some idea on performance.
2. What is the minimal TPU required? 16x16=256 seems to be overkill for a 109B model.
3. I am still curious on why tp+fsdp (without expert parallel) leads to OOM? Do you have more insights?

Issue #14, opened 2025-01-08
Remove dependency on HuggingFace token


tengyifei:
Right now we need a HuggingFace token in order to run tests/run the model locally. This token is used to fetch the model config and fetch tokenizers.

It should be possible to drop in a tokenizer model as a file. It should also be possible to specify a model config as a `dict` as opposed to a `str` like `Meta-Llama/Llama-3.1-8B`, which would cause `transformers` to look up the model using a token. Once we can do these, we'll no longer need a token to do dev work and testing.

PR #357, opened 2025-08-05
Remove HF_TOKEN dependency in E2E test

jack8558:
Removing HF_TOKEN dependency in E2E test

- Created `tp save_hf_model_files_to_gcs` to save huggingface model files in gcs
- Saved tokenizers and llama3-8B's model weights and configs  in gcs [bucket](http://shortn/_YaYVZG9n4b) (the weights and configs are needed for SFT e2e test)
- Since huggingface library can't load directly from gcs, added util funciton `copy_gcs_to_local` which download in tmp directory
- Removed HF_TOKEN on e2e test and cpu test

`tp save_hf_model_files_to_gcs` example:
```
tp save-hf-model-files-to-gcs \
  --repo-id "meta-llama/Meta-Llama-3-8B" \
  --gcs-path "gs://bucket" \
  --file-type "all" \
  --temp-dir /mnt/disks/tmp
```

https://github.com/AI-Hypercomputer/torchprime/issues/14


.github/workflows/e2e_test.yml
```diff
@@ -78,6 +77,7 @@ jobs:
             --name $name \
             torchprime/torch_xla_models/train.py \
             model=llama-3-8b \
+            model.tokenizer_name=gs://torchprime/e2e-test/hf-model-files/meta-llama-3-8b \
```
vlasenkoalexey, line 80: could you please check that this folder is publicly accessbile

torchprime/torch_xla_models/model/base_causal_lm.py
```diff
@@ -153,8 +155,13 @@ def _maybe_save_checkpoint(self, config: DictConfig) -> None:
     # Step 3: Save the HF config files and tokenizer
     if xr.process_index() == 0:
       logger.info("Saving Hugging Face configs and tokenizer to %s", save_dir)
-      model_utils.copy_hf_config_files(config.model.pretrained_model, save_dir)
-      model_utils.save_hf_tokenizer(config.model.pretrained_model, save_dir)
+      # Copy to local if in GCS
```
vlasenkoalexey, line 158: Could you explain why is it necessary?
If training started from gcs fuse which looks like a local folder, would it still try to copy?

torchprime/torch_xla_models/model/model_utils.py
```diff
@@ -484,3 +500,53 @@ def save_hf_tokenizer(model_path_or_repo: str, save_dir: Path) -> None:
...
+    return path_or_repo
+
+  if not shutil.which("gsutil"):
+    raise RuntimeError(
+      "gsutil command not found, but is required for downloading from GCS. "
+      "Please install the Google Cloud SDK."
+    )
+
+  local_dir = tempfile.mkdtemp()
+  _TEMP_DIRS_TO_CLEAN.append(local_dir)
```
vlasenkoalexey, line 533: This is a bad pattern, could you make temp dir as an argument or use context manager to auto clean it?
If that's inconvenient, feel free to leave it as is.

vlasenkoalexey (changes requested):

.github/workflows/e2e_test.yml
jack8558: The gs://torchprime bucket is not publicly accessible: http://shortn/_wmSPjtpxwl

Only users who are authenticated will be able to access it.

torchprime/torch_xla_models/model/model_utils.py
jack8558: Updated the function with context managager. Lmk if this look better.

torchprime/launcher/cli.py
```diff
@@ -77,6 +79,54 @@ def cli(ctx, interactive):
   ctx.obj["interactive"] = interactive
 
 
+@cli.command("save-hf-model-files-to-gcs")
```
vlasenkoalexey, line 82: probably you should submit your PR for removing click dependency first and then update this one

torchprime/launcher/save_hf_tokenizer_and_model.py
```diff
@@ -0,0 +1,94 @@
...
+import os
+import subprocess
+import tempfile
+from pathlib import Path
+
+from huggingface_hub import snapshot_download
+
+logger = logging.getLogger(__name__)
+
+TOKENIZER_PATTERNS = [
```
vlasenkoalexey, line 13: nit for constants it is recommended to use immutable data types, tuple in this case:
TOKENIZER_PATTERNS = (a, b, c,)

vlasenkoalexey (approved):

PR #365, opened 2025-08-20
Able to run torchax with llama model

jialei777:
Fixed issues:
- reduce batch size in `default.yaml` to be able to run at 4-chip vm
- add `is_view_op` arg to `torchax.ops.ops_registry.Operator` class
- 10x smaller parameter init to avoid overflow
- add `torchax.enable_globally()` in the beginning
- fix an issue related to mask in splash attention
- fix issue when `step_compiled.cost_analysis()` returns a dict instead of a list
-  add gradient clip

run training:
> python torchprime/experimental/torchax_models/run.py model_impl=orig

### DIFF
--- a/torchprime/experimental/torchax_models/run.py
+++ b/torchprime/experimental/torchax_models/run.py
@@ -146,6 +146,7 @@ def register_attention(fn):
 def make_weight_shard(weight_meta, slice_index):
   weight_shard_meta = weight_meta[slice_index]
   with torchax.default_env():
+    # TODO(jialei): update to better weight init methods
     return interop.jax_view(
       torch.randn(weight_shard_meta.shape, dtype=weight_shard_meta.dtype) * 0.1
     )
--- a/torchprime/experimental/torchax_models/splash_attn.py
+++ b/torchprime/experimental/torchax_models/splash_attn.py
@@ -63,6 +63,7 @@ def tpu_splash_attention(
 
     if mask is None:
       mask = splash_attention_mask.CausalMask(shape=(query.shape[2], query.shape[2]))
+
     # Create multi-head mask
     multi_head_mask = splash_attention_mask.MultiHeadMask(
       masks=(mask,) * query.shape[1]
--- a/torchprime/experimental/torchax_models/configs/default.yaml
+++ b/torchprime/experimental/torchax_models/configs/default.yaml
@@ -1,7 +1,5 @@
 # The default config file. You may override configs with `key=value` arguments on the CLI
 # according to https://hydra.cc/docs/advanced/override_grammar/basic/.
-
-
 global_batch_size: 8 # runable on a v6e-4 vm
 model_type: 8B
 lr: 5.e-5


torchprime/experimental/torchax_models/train.py
```diff
@@ -164,8 +167,10 @@ def train_loop(
   env = torchax.default_env()
 
   jax_params = env.t2j_iso(weights)
-  # jax_optimizer = optax.adamw(lr)
-  jax_optimizer = optax.sgd(lr)
+  jax_optimizer = optax.chain(
+    optax.clip_by_global_norm(1.0),  # Clip gradients to a max norm of 1.0
```
vlasenkoalexey, line 171: could you move this to config or at least to the constant

torchprime/experimental/torchax_models/splash_attn.py
```diff
@@ -63,14 +63,17 @@ def wrap_flash_attention(mask, query, key, value, decoder_segment_ids):
       v_layout=splash_attention_kernel.QKVLayout[global_v_layout],
     )
 
-    if mask is not None:
-      mask = splash_attention_mask.CausalMask(shape=(query.shape[2], query.shape[2]))
-
+    # FIX: If the input mask is None, create a default CausalMask.
```
vlasenkoalexey, line 66: sorry, I'm not following logic here.
Code used to have check for mask creation, why are you removing that code and adding a comment to fix it?

torchprime/experimental/torchax_models/splash_attn.py
```diff
@@ -24,7 +24,7 @@ def tpu_splash_attention(
       decoder_segment_ids, decoder_segment_ids
     )
 
-  print("HERE", locals())
+  # print("[DEBUG][tpu_splash_attention] --->", locals())
```
vlasenkoalexey, line 27: please remove

torchprime/experimental/torchax_models/run.py
```diff
@@ -134,15 +134,20 @@ def register_attention(fn):
...
+    is_view_op=False,
   )
 
 
 def make_weight_shard(weight_meta, slice_index):
   weight_shard_meta = weight_meta[slice_index]
   with torchax.default_env():
     return interop.jax_view(
-      torch.randn(weight_shard_meta.shape, dtype=weight_shard_meta.dtype)
+      torch.randn(weight_shard_meta.shape, dtype=weight_shard_meta.dtype) * 0.1
```
vlasenkoalexey, line 150: just a question, should it be xavier or kaiming instead of randn?

vlasenkoalexey (approved):

torchprime/experimental/torchax_models/splash_attn.py
jialei777: Previous code add mask if it is not None. seems to be a bug?

torchprime/experimental/torchax_models/run.py
jialei777: probably, adding *0.1 fixed the overflow issue so i just add that simple fix.

PR #370, opened 2025-08-26
Update MFU computation to align with JAX (MaxText)

jialei777:


torchprime/metrics/tests/test_mfu.py
```diff
@@ -1,151 +1,68 @@
+import sys
+from pathlib import Path
+
 import pytest
+import yaml
 
 from torchprime.metrics.mfu import compute_mfu
 
+ROOT = Path(__file__).resolve().parents[3]
```
vlasenkoalexey, line 9: This looks terribly hacky, could you check if hydra+omegaconf could be leveraged for this?
As far as I understand hydra has tools to be able to find configs located at right place

vlasenkoalexey (approved):

PR #382, opened 2025-09-03
Fix backward mismatch in deepseek

jialei777:
Main change:
- Make precision back to fp32 (previously change all of them to bf16 with the hope to improve performance)
- convert the layer/state_dict back from grouped MoE (i.e., stack experts to 1 tensor instead of multiple tensors) to regular HF version of MoE.
- change the testing logic: when doing grad comparison, we need to first convert the model state dict to ensure the layer/tensor name match HF

### DIFF
--- a/torchprime/torch_xla_models/model/deepseek_v3/model.py
+++ b/torchprime/torch_xla_models/model/deepseek_v3/model.py
@@ -685,7 +685,7 @@ def revert_grouped_moe_to_hf_state_dict(
 
   Args:
     grouped_state_dict (dict): The state_dict with keys like '...grouped.W_gate'.
-    config: The model configuration, used to get the number of experts.
+    n_routed_experts: The number of experts used in the model.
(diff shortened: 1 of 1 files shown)


torchprime/torch_xla_models/model/deepseek_v3/model.py
```diff
@@ -274,13 +274,13 @@ def _post_state_dict_old_keys(self, state_dict, prefix: str):
     E = self.E
     for e in range(E):
       state_dict[f"{prefix}experts.{e}.gate_proj.weight"] = (
-        self.grouped.W_gate[e].t().contiguous().to(BF16)
+        self.grouped.W_gate[e].t().contiguous().to(FP32)
```
vlasenkoalexey, line 277: was it FP32 in HF original implementation?

I understand that layernorms can be FP32, but MOE weights are largest tensors in the whole model, they probably should be bf16.

jack8558:
Thank you for fixing this!

Issue #377, opened 2025-09-02
Running the test_run_model test results in a segmentation fault error


erichuang-cienet:
Environment:
- TPU VM: v6e-8
- Python 3.11
- torch 2.9.0.dev20250901+cpu
- torch-xla 2.9.0+git4c586bd

Command:
```
export LIBTPU_INIT_ARGS='--xla_tpu_use_enhanced_launch_barrier=false'
pytest -v torchprime/launcher/test_run_model.py
```

Error:
```
torchprime/launcher/test_run_model.py::test_run_model_transact Fatal Python error: Segmentation fault

Thread 0x000076568faac800 (most recent call first):
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/torch_xla/core/xla_model.py", line 234 in unlazy
  File "/home/sa_106246016191652103077/torchprime/torchprime/launcher/run_model.py", line 45 in run_model_xla
  File "/home/sa_106246016191652103077/torchprime/torchprime/launcher/test_run_model.py", line 29 in test_run_model_transact
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/pluggy/_callers.py", line 121 in _multicall
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/pluggy/_hooks.py", line 512 in __call__
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/_pytest/python.py", line 1627 in runtest
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/_pytest/runner.py", line 174 in pytest_runtest_call
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/pluggy/_callers.py", line 121 in _multicall
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/sa_106246016191652103077/torchprime/venv/lib/python3.11/site-packages/pluggy/_hooks.py", line 512 in __call__
[...]

PR #387, opened 2025-09-09
Fix test_run_model from hanging when running on TPU

jack8558:
When running test_run_model on TPU, it was hanging indefinitely. I realized that the hanging was occuring when testing the torchax part and when I divide them to two separate test, hanging stopped.

I am not exactly sure why this is happening but my best guess is some conflict between initializing backend for TPU. If anyone knows better explanation, please let me know.

Environment:
- TPU VM: v4-8
- Python 3.11

### DIFF
--- a/torchprime/launcher/test_run_model.py
+++ b/torchprime/launcher/test_run_model.py
@@ -12,22 +12,32 @@ class SimpleModel(torch.nn.Module):
     return (torch.randn(batch_size, 10),), {}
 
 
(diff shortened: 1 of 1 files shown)


erichuang-cienet:
I also ran into this issue. I was able to workaround it by disabling the `xla_tpu_use_enhanced_launch_barrier` flag.  #377

jialei777 (approved):

